PR #32, opened 2017-06-22
close #23

### LINKED ISSUE #23: Color Extraction when using live wallpaper
https://stackoverflow.com/questions/23294489/android-convert-current-screen-to-bitmap

YAJATapps:


deletescape:
Looks good, will merge now, as it won't break any current feature atleast

Issue #38, opened 2017-06-26
Implement HockeyApp/TestFairy for testers


fonix232:
I know it is a very low priority task, but it would be nice if builds from Travis would be distributed through an automatic way.

Since Travis supports uploading even non-tagged builds to pretty much any such tool, and I have some experience with HockeyApp, I'd gladly implement it as long as it is okay for the main devs.

It would also mean a bit of branching here and there, with `master` being used only for version-tagged releases, and a separate `beta` and `develop` branch, all three pushing APKs to different HockeyApp configs, thus establishing the alpha, beta and release channels.

deletescape:
If you could provide the structure for this in a PR I would really appreciate it! Never worked with Hockey before, but sounds like a great idea!

PR #45, opened 2017-06-27
Added HockeyApp distribution script

fonix232:
Do not merge until `HOCKEYAPP_TOKEN` is set in Travis, otherwise it *will* cause unsuccessful deployments.


fonix232:
Fixes #38

PR #46, opened 2017-06-27
Move upload to deployment

fonix232:
This will avoid uploading APKs from pull requests. Should probably move


fonix232:
DO NOT MERGE YET

PR #130, opened 2017-07-05
Support generating replacement icons from icon packs

paphonbth:


edwaine:
I hope you can also include support for old "Google Camera" app to Pixel themed icon.

Issue #260, opened 2017-07-20
"Event not found" toast when opening GCalendar via clock


Chaphasilor:
Everytime I click on the clock in the Google too bar, the Calendar opens normally, but there is a toast popup saying "Event not found". Maybe it's calling the wrong activity?

G0ldenSp00n-zz:
I fixed this bug and issued a pull request.  Hopefully should be fixed by next release.

AxiK101:
Nope, not fixed for me. Xperia XA Ultra, 7.0 stock android

G0ldenSp00n-zz:
You have to wait till it is accepted by the author, and then you might have to wait till a patch is pushed.

suphon-t:
Fixed with #274

PR #274, opened 2017-07-21
Fixed Bug with DateWidget

G0ldenSp00n-zz:
Fixed the DateWidget onClick intent code, it will now no longer produce a toast notification when clicked.  Also has a fallback if running Android below Ice Cream Sandwhich. Fixes bug #260

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -24,7 +24,7 @@
       </value>
     </option>
   </component>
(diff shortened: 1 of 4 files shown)


suphon-t:
Please remove your modules.xml changes.
Also you don't need the else part as this launcher has minApi 21

G0ldenSp00n-zz:
Removed module.xml changes.

AxiK101:
Nope, not fixed for me. Xperia XA Ultra, 7.0 stock android

PR #294, opened 2017-07-24
Italian strings improvements

mj1687:
This edit don't radically change the translations but make them bit more "italian" and friendly (i think)


nyancrimew:
Translation changes should be done on crowdin only

Issue #309, opened 2017-07-26
Icon and Icon text scale resetting to 98%


uzair-inamdar:
This only happens when you set these scales to 100%, after which they automatically go back to 98% on going back and reopening the UI setting.

hassie-dash:
Essentially what is happening is that the rounding is causing some numbers to be missed. I don't think there is a way to fix that, unless we keep the range between the minimum and maximum at 100.

![screenshot 54](https://user-images.githubusercontent.com/26469286/28619230-4b3bf8c6-71ff-11e7-9b5b-c323b3543e2c.png)

PR #315, opened 2017-07-26
Fix for Issue #309

hassie-dash:
Changed minimum seekbar values to 0.5 to keep range at 100.


nyancrimew:
This isn't really a fix, but I know how to properly fix it, will be done as soon as I have time to do it

PR #324, opened 2017-07-27
Rewrote Open Weather Map weather downloader

hassie-dash:
* Rewrote the weather downloader for Open Weather Map, dependency no
longer required.
* Added hidden preference where users can now enter their own API key.

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/QsbBlockerView.java
+++ b/app/src/main/java/ch/deletescape/lawnchair/QsbBlockerView.java
@@ -24,6 +24,10 @@ import ch.deletescape.lawnchair.pixelify.OnWeatherInfoListener;
 import ch.deletescape.lawnchair.pixelify.ShadowHostView;
 import ch.deletescape.lawnchair.pixelify.WeatherInfo;
 import ch.deletescape.lawnchair.pixelify.WeatherThing;
+import ch.deletescape.lawnchair.weather.WeatherUnits;
+import ch.deletescape.lawnchair.weather.owm.OWMWeatherCallback;
+import ch.deletescape.lawnchair.weather.owm.OWMWeatherData;
+import ch.deletescape.lawnchair.weather.owm.OWMWeatherDownload;
 
 public class QsbBlockerView extends FrameLayout implements Workspace.OnStateChangeListener, OnWeatherInfoListener {
     public static final Property<QsbBlockerView, Integer> QSB_BLOCKER_VIEW_ALPHA = new QsbBlockerViewAlpha(Integer.TYPE, "bgAlpha");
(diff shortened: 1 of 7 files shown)


nyancrimew:
@fonix232 @paphonb opinions on this? Imo we can merge

app/src/main/java/ch/deletescape/lawnchair/weather/owm/OWMWeatherDownload.java
```diff
@@ -0,0 +1,100 @@
...
+    private JSONObject download() {
+
+        String mURL;
+
+        if (weatherUnits == WeatherUnits.IMPERIAL)
+            mURL = "http://api.openweathermap.org/data/2.5/weather?q=" + location + "&units=imperial&APPID="
+                    + apiKey;
+        else
+            mURL = "http://api.openweathermap.org/data/2.5/weather?q=" + location + "&units=metric&APPID="
+                    + apiKey;
```
00raq00: if (weatherUnits == WeatherUnits.IMPERIAL)
             mURL = "http://api.openweathermap.org/data/2.5/weather?q=" + location + "&units=imperial&APPID="
                   + apiKey;
         else
            mURL = "http://api.openweathermap.org/data/2.5/weather?q=" + location + "&units=metric&APPID="
                    + apiKey;

i think better it will be look like 

`   mURL = "http://api.openweathermap.org/data/2.5/weather?q=" + location + "&units="+weatherUnits +"&APPID="
                   + apiKey;`

app/src/main/java/ch/deletescape/lawnchair/QsbBlockerView.java
```diff
@@ -136,18 +130,25 @@ public void onClick(View view) {
...
-                    helper.getCurrentWeatherByCityName(city, callback);
+                    SharedPreferences prefs = Utilities.getPrefs(getContext());
+                    String apiKey = prefs.getString("pref_weatherDebug_api_key", BuildConfig.OPENWEATHERMAP_KEY);
+                    String city = prefs.getString("pref_weatherDebug_city", "Lucerne, CH");
+                    String units = prefs.getString("pref_weatherDebug_units", "metric");
+                    final boolean isImperial = units.equals("imperial");
+                    new OWMWeatherDownload()
+                            .setAPIKey(apiKey)
+                            .setLocation(city)
+                            .setWeatherUnits(isImperial ? WeatherUnits.IMPERIAL : WeatherUnits.METRIC)
```
00raq00, line 141: here u can use setWeatherUnits(units) instead of check if units is equeal "imperial" and then set again check if isImperial then use WeatherUnits.IMPERIAL ...

00raq00:
i think better use published lib (com.github.KwabenBerko:OpenWeatherMap) if it works. 
In hassie-dash code is mess. He concat string url instead of use lib to odata or http get

hassie-dash:
Switched to URI builder in latest commit.
```java
Uri uri = Uri.parse("http://api.openweathermap.org/data/2.5/weather?")
 +                    .buildUpon()
 +                    .appendQueryParameter("q", location)
 +                    .appendQueryParameter("units", weatherUnits == WeatherUnits.IMPERIAL ? "imperial" : "metric")
 +                    .appendQueryParameter("APPID", apiKey)
 +                    .build();
 +            URL url = new URL(uri.toString());
              HttpURLConnection request = (HttpURLConnection) url.openConnection();
```

app/src/main/java/ch/deletescape/lawnchair/QsbBlockerView.java
suphon-t: No you can't. `units` is a `String` but the method argument requires a `WeatherUnits`.

suphon-t:
@deletescape I think the library implementation is already good enough. But this is just my opinion.

hassie-dash:
My reasoning was to remove the un-necessary parsing of the whole json object and just extract the data required for better performance, but if the library suits more, that's no issue 😊

PR #336, opened 2017-07-28
Lawnchair: use simple set wallpaper chooser

martincz:
Show all apps that support it and let user decide is a better way.
If the user haven't install Google WallPaperPicker, the current logic will cause the application to force a stop.

Change-Id: Ied043315a07a530aaf1a272b415c2bd5864d8512


nyancrimew:
Current implementation does not cause to force stop without google wallpapers, instead it will fallback to the built in wallpaper chooser

martincz:
Hmm... You are right. It's only happened when package name changed.
MoKee's Lawnchair use a different package name to avoid conflicts with the original version.
If you want, you can cherry-pick this commit to optimize logic.
https://review.mfunz.com/#/c/22537/

PR #342, opened 2017-07-29
App drawer order of apps

Williwood:


nyancrimew:
This is not how this works, please only make PR's on the alpha branch and do only include new commits.

PR #371, opened 2017-07-31
Lawnchair: set city value to preference summary if it is defined.

martincz:
Change-Id: I138e738f3c5f9c6df63977a4c92c354d20651d09


martincz:
about new icon picker layout
old:(http://imgur.com/nRWvN7q)
new:(http://imgur.com/JlQF2aK)

PR #448, opened 2017-08-08
Lawnchair: set a demo api key

martincz:
Set a demo api key as a user reference.

Change-Id: Ib6d33886d2faaa46754f132d57cb6f52bc00f522


nyancrimew:
Just be aware that this key might be blocked if too many user make requests with it

PR #469, opened 2017-08-10
Refactoring and extending preferences

MFlisar:
What I did is following:

* extracted all strings that are saved to preferences and stored them in on single file: `PreferenceFlags`
* extracted setting prefix strings as well, the new `IPreferenceProvider` must handle them instead of the function caller
* all default strings like the default weather city are also defined in `PreferenceFlags` now
* created an interface that implements the settings: `IPreferenceProvider`; this one is used EVERYWHERE but in the preference fragments, this allows a easy extension of the settings setup views (i.e. if the app should be integrated in another app and many preferences may already exist in this app and should be derived from the app)
* added new feature: allow to use the animated clock icon for non stock clock apps, currently only samsung clock is supported, because I don't know other apps package and activity names; by default this is turned off for consistancy

Result:

* `FeatureFlags.INSTANCE.<function>` and `FeatureFlags.<function>` are replaced by `Utilities.getPrefs(context).<function>`

Further improvements:

* `Utilities.getPrefs(context).<function>` => maybe with kotlin this could be somehow easily replaced with something like `Pref.<function>`? Did not check this, as kotlin is new to me
* the new implementation easily allows to use a single `SharedPreference` instance in the whole app, anything that speaks against this? `SharedPreference` is even thread save as far as I know, not sure if there does speak anything against this


MFlisar:
* changed the provider implementation to use the dynamic `$applicationId` instead of hard coded application id prefixes
* add a themer that should take care of any theming functions in the future, started with the drawer background and blur color

PR #587, opened 2017-08-23
Opti.sh: Image compressor and optimizer

HiFIi:
Never hurts :)


nyancrimew:
All image assets in Lawnchair are optimized already. And this PR alone doesn't do anything in itself except adding a shell script.

Issue #619, opened 2017-08-26
App info on home screen for apps dragged from the apps drawer


theandroidenthusiast:
Can you please put in the app info button between cancel and uninstall when you drag a app from the drawer? It just is second nature to grab it from the app drawer and go into app info settings. Thank you

PR #660, opened 2017-09-02
Add app info drop for apps from drawer

ghost:
Not tested
Issue #619


LinuxLlama:
It would be nice if a toggle was added so if you don't like the feature, you can disable it.

suphon-t:
And you already have app info in the long press shortcuts too.

ghost:
In fact, it is a feature that we find on the Pixel Launcher and that was implemented only half in this launcher. It was working from the home screen but not from the app drawer. You must enable developer options to see it. I agree that it is a bit redundant but some seem to prefer it.

PR #704, opened 2017-09-10
Make round "Search apps" text use accent color

LinuxLlama:
Just a simple fix to change the white text to blue.

screenshot sep 10 2017 11-03-27 am


suphon-t (changes requested):
It is intended to be white on black / grey on white. Please make this an option. Also, the accent color is not always blue so please change from a hardcoded color to dynamic accent color instead.

PR #744, opened 2017-09-23
Allow icon labels to use 2 lines

### LINKED ISSUE #301: App Drawer Icon Label Truncated
For apps with longer label (app name) it truncates the text (check the picture for the Samsung... apps) . Can it be made to split the long text in 2 lines? 
screenshot_20170725-184438

Rawi666:
This fixes #301 

Before
before

After
after


LinuxLlama:
Could you possibly have an option to toggle in behavior settings?

Rawi666:
Sure, I'll do that. I just thought that this was a bug/an oversight rather than an intended feature :)

Rawi666:
Done. Let me know if there is something more to change in this commit

PR #749, opened 2017-09-24
Fix for icon shape 'Do not change' not working properly

### LINKED ISSUE #661: Issue with icon shapes and apps that can change their icons
Since I updated to 1.1.0.1233
- Combine Pixel style preference to icon shape on API24+

G Calendar switched to a round icon while all the other icons stayed the same.

I tried messing with the shape preferences, but "Do not change" doesn't seem to work properly anymore. It just goes to random circles here and there or retains the previous shape.

"None" breaks app that can change their icon. Basically when I try to edit and icon on the homescreen or drawer, it will be shown correctly but the one that ends up being displayed isn't the correct one (apps : flamingo, snapcamera HDR)

Right now I messed alot with shapes, rebuilt database multiple times, thankfully I was able to get rid of the circles except for Flamingo which default to white circle background, and snap (red gallery instead of blue camera, for some reason)

Device is Bacon/A0001 , 7.1.2 AOSP+CAF

Lawnchair build 1238

Rawi666:
This fixes #661 
After "Combine Pixel style preference to icon shape on API24+" commit, pixel style override "Do not change" and "Circle" was the same.
Now its as follow:
Do not change - use AOSP icons but with original shapes
None - use system icons (on my Huawei P10 it shows ugly Huawei icons so its fine)
Circle - use AOSP in circle

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/graphics/IconShapeOverride.kt
+++ b/app/src/main/java/ch/deletescape/lawnchair/graphics/IconShapeOverride.kt
@@ -47,7 +47,6 @@ class IconShapeOverride {
     companion object {
 
         const val planeMask = "M21,16V14L13,9V3.5A1.5,1.5 0 0,0 11.5,2A1.5,1.5 0 0,0 10,3.5V9L2,14V16L10,13.5V19L8,20.5V22L11.5,21L15,22V20.5L13,19V13.5L21,16Z"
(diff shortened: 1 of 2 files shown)


nyancrimew:
I like this, thanks a lot

Younes-L:
Thank you !

suphon-t:
The intended behavior of "Do not change" is to use system settings (circle if the system choose round icons and normal icons otherwise). And the none option is for icons to have freeform shape.

PR #751, opened 2017-09-25
Icon text scalling fixes

### LINKED ISSUE #750: 2lines app titles get cropped
**BUG** 
In app drawer when enabling the new 2line style for app titles, the second line in some characters is getting cropped. Grid size is set in auto for both columns and rows. Only when changing to manual row number this is fixed BUT then you get a pretty big space between rows. In desktop is ok.

**DEVICE SPECS:** 
Sony Xperia XA Ultra, Stock 7.0
Lawnchair 1.1.0.1265

Rawi666:
Fixes not working app drawer text scalling (typo) and increases padding a bit if 2 lines labels are enabled.
Fixes #750 (at least on 6" emulator in 1080p)


nyancrimew:
Exactly the fix I wanted to implement, thanks a lot!

PR #756, opened 2017-09-25
Add option in preferences to animate alternative clock apps

divadsn:
This should finally allow users to enable this "hidden" option when they're using stock OEM apps.
Now it should animate the clock app on Samsung and Xiaomi devices as well.

TODO: Find a regex or way to add all clock apps (maybe by looking up clock providers?)


divadsn:
Sorry @deletescape that I added two more strings to translate 😛

Issue #759, opened 2017-09-25
Change padding size on dock


LinuxLlama:
If somebody asks for a feature request on Google+, I'll post them here from now on. This has been forwarded from Luke Stonbraker on Google+: 

Could you add an option to reduce the amount of padding and icon size on the dock?

nyancrimew:
The dock icon size can be adjusted already

LinuxLlama:
@deletescape I knew that. I just copied the full message from Google+.

PR #766, opened 2017-09-27
Added Dock Height Scale slider in preferences

Rawi666:
This is not necessary for most phones but there are some situations like my wife's Nexus 5 which is designed to have a 4x4 grid. If I set the grid manually to 4x5 the dock takes too much space and it's looking a bit weird. Usually I have a custom build for this phone with different height but this will be helpful for others.
This was also requested on #759 and there was somewhere a bug where if you scale dock icons up/down then the dock height is not adjusted accordingly so this additional slider can help in both cases.


divadsn:
It's not intended to be rude, but it would be nice to see your previous issues solved before fixing other things :)

https://github.com/Deletescape-Media/Lawnchair/issues/750

Rawi666:
For #750 I have an idea with one more slider in preference ;) It's a beauty of Android where every phone works different

PR #806, opened 2017-10-05
F-Droid build flavor

### LINKED ISSUE #240: Add to F-Droid repository
Hi, keeping the app up to date would be very easy if you submit it to f-droid: https://gitlab.com/fdroid/rfp/issues

"F-Droid is an installable catalogue of FOSS (Free and Open Source Software) applications for the Android platform. The client makes it easy to browse, install, and keep track of updates on your device."

dimqua:
Fixes #240.


guest4711:
Thanks, looks good...

fonix232 (approved):
Looks good to me.

nyancrimew:
Looks good to me too

nyancrimew (approved):

suphon-t (changes requested):
- `WeatherAPI.kt` in `fdroid` could be moved to main with the one that is already in `play` variant overriding it (Which might be moved to resource overrides with reflections instead)
- Instead of creating a new preference file, we could move the default value into an `int` resource and override that resource instead. This will help in case we need to add something to that preference.

Otherwise it looks good to me

divadsn (changes requested):
As @paphonb already mentioned, I think there should rather be one preferences file for both flavors, it will make things simpler for future PRs and changes.

I think it should be possible by hide the options which require Google services for the F-Droid flavor.

fonix232:
@dimqua once you add said fixes, I think merging can go on and we can release on F-Droid ASAP.

dimqua:
>there should rather be one preferences file for both flavors

Sorry, but I don't know how to do this right.

suphon-t:
> Sorry, but I don't know how to do this right.

As I said, move the default to a separate resource and override that resource instead.

dimqua:
I can't move _WeatherAPI.kt_ to `main`, since in this case I [duplicate](https://travis-ci.org/Deletescape-Media/Lawnchair/builds/286632246#L1615) class _WeatherAPI_ in the same file in `play`.

divadsn:
@dimqua could you please maybe upstream your repo and apply your changes to our new build.gradle, as we did a lot of changes to get the app ready for the Play Store, so it should be now easier. :hugs:

dimqua:
@divadsn Yes, but according to GitHub this PR "has no conflicts with the base branch".

divadsn:
@dimqua yes, but we also have changed a lot in build types and the F-droid release doesn't need to be a debug apk, right? As we won't be displaying the Google Now Feed :wink:

dimqua:
AFAIK, F-Droid doesn't build debug build types anyway. Is there something I should add to fdroid build flavor specifically? 
I assume to "ENABLE_LAWNFEED" should always be "false" for F-Droid release.

nyancrimew:
ENABLE_LAWNFEED is supposed to be true for release builds as without it the swipe for google now will NEVER work, with it enabled it would be possible to make a custom Lawnfeed build to specifically work with the fdroid release.

nyancrimew:
@divadsn do we need changes to our travis script so the right flavour is built on our builds?

divadsn:
@deletescape no afaik, haven't tested it out so far as a lot of changes happened recently.

ildar:
I am sorry, I'm a little lost. What's the current status of this PR?

divadsn:
It's currently on hold as relevant changes are still being made, I will
finish the flavor after we rolled out the next Play Store release.

fonix232:
@divadsn what is your status on this?

ghost:
I'd like Lawnchair on F-droid for xmas  this year! How do the odds look for that to happen?

nyancrimew:
@haverholm I'd like the Pixel 2 and 5 bitcoins for xmas this year!

Jk, we'll see what we can do :)

ghost:
Yeah, those sound good too! Here's hoping we both get our wishes ;)

ildar:
I'm very sorry, I don't fully understand github staff. Is it correct that
we now have 2 positives, 2 negatives and failing CI build?

TheComputerGuy96:
How about adding an option to use DuckDuckGo (while changing the icon in the built-in search button in the homescreen as well)?

divadsn:
As the PR has a lot of work to do and has lost the track of recent changes I will close this PR and re-open as a new one later.

ildar:
Excuse me, @divadsn , are you going to take over or @dimqua will continue?

divadsn:
@ildar Take over in another PR, will cherry-pick it.

ildar:
I see. Thanks. I'll follow up.

PR #813, opened 2017-10-08
Unicode insensitive comparison in All Apps search

Rawi666:
For searching replace any non-ascii characters to theirs ascii equivalents.
E.g. in Polish: "Menadżer telefonu" can now be found using "menadzer" or "menadżer" or "dzer" or "dżer"

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -91,6 +91,7 @@ dependencies {
     implementation 'com.google.android.gms:play-services-awareness:11.2.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
     implementation 'com.pavelsikun:vintage-chroma:1.5'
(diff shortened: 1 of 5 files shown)


divadsn (changes requested):
I don't like the way you replace the speicial unicode characters (umlauts) using String#replace for every you have found, I think there is a better and more effective way to achieve this.

Edit: Actually there is a better way, have you tried that?
https://stackoverflow.com/a/3322174

divadsn:
As we came to a decision, there is no need in my opinion to replace Greek and Russian letters, as they have their own keyboard layouts, the original idea was also to avoid SMS size limitations, while we don't have such limitations when it comes to the search bar.

divadsn (approved):

divadsn:
There is also actually no need to add a entire library for one usage :smile:

PR #840, opened 2017-10-14
Bring back changelog to default build config

divadsn:


divadsn:
Merge that after fixing s3-upload.sh!

PR #867, opened 2017-10-21
Move Color/Integer definitions from styles.xml to color.xml/integers.xml

nschnettler:
This commit makes it easier for substratum themers to theme this app. It is already themeable by overlaying the styles, but moving the resource definitions to their specific files makes basic theming possible without touching the styles.

Theming colors.xml is less "invasive" than theming the styles. Example: My current Overlay using a style overlay works just fine, but after an update of LawnChair there is a chance, that the app stopps working until the user rebuilds the Overlay. Using just a colors Overlay this does not happen.

### DIFF
--- a/app/src/main/res/values-v23/styles.xml
+++ b/app/src/main/res/values-v23/styles.xml
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <style name="SettingsTheme" parent="@style/Theme.AppCompat.Light">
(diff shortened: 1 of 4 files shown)


nyancrimew (approved):
Looks good to me

divadsn (commented):
Will approve after final test results tomorrow.

divadsn:
I will test it with different themes now to see if there are some issues eventually.

Also, some conflicts needs to be resolved tho.

divadsn:
Gonna cherry-pick that, as @paphonb have broken the branch with his merge 😆

PR #883, opened 2017-10-27
build.gradle: use implementation instead of compile

chinmaykunkikar:
Fix: "Configuration 'compile' in project ':launcherclient' is deprecated. Use 'implementation' instead."


divadsn:
This has been already fixed, it's just not pushed yet.

PR #889, opened 2017-10-29
Dynamic status bar gradient by luminance calculation

saikrishna279:
This is like a switch to turn ON/OFF the annoying white gradient over the status bar area when a dark/black wallpaper is applied. Especially noticeable on LED Displays
It is replaced by a fully transparent background when a dark/black wall is applied
The gradient is shown when a light/bright wallpaper is applied

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/Launcher.java
+++ b/app/src/main/java/ch/deletescape/lawnchair/Launcher.java
@@ -28,6 +28,7 @@ import android.app.ActivityOptions;
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.app.SearchManager;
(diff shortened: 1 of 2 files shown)


nyancrimew:
As we have discussed on Telegram already this white gradient can currently be turned off manually with a switch, but personally think it would be great to replace said switch with your automatic method. But this would require some adjustments to your PR (like removing the switch, etc.)

divadsn:
Imo it's better with a switch, for live wallpapers especially.

divadsn:
As I already mentioned, it would break with live wallpapers, same as with blur.

AxiK101:
Switch indeed

PR #891, opened 2017-10-30
backuprestore: rebuild icon db after every Settings restore

chinmaykunkikar:
The IconCache doesn't clear itself even after killing the Launcher process, leaving the icon shapes unchanged.
Rebuilding icon db fixes it.


divadsn:
I'm not sure if it will cause the same issue we had before fixing Oreo support...

chinmaykunkikar:
Messed up branching while submitting a seperate PR for the last commit.
Any way to fix it?

nyancrimew:
Locally reset the history and force push this branch, the PR will be updated.

Now about the PR: @divadsn this should technically be fine/desired, as we also rebuild the db after any of the iconpack related settings change, so I guess this makes sense here.

nyancrimew (approved):

PR #897, opened 2017-11-03
Less smiles = more smiling devs

clairekardas:


YUDHPK:
Why did you remove access to some apps.
It isn't a positive work.(i should be able to control which apps i need/use not someone else)
It forced me to rethink my decision of making your awesome launcher the default one.
I recommended lawn chair to many of my friends because i liked it's focus on constant development.
Please revert it
@deletescape 

Issue #897
[emoji1308]

mbirth:
It's a launcher. The main purpose is to show ALL installed apps and be able to *launch* them. This is unacceptable. Besides what those apps do, I want a launcher that doesn't censor what I am able to see. What's next? Hiding malware apps from me so they can run in the background without me noticing that they are there?

amirzaidi:
This is pretty damn stupid.

guest4711:
Thank you for the list of apps I should try. I had never heard of them before this commit ;-)

PR #901, opened 2017-11-05
#MakeLawnchairGreatAgain

ghost:
You know what you did.


KieronQuinn:
Without testing this, I believe this would break the entire launcher. startsWith("") returns true for **every package name** (and [this](https://stackoverflow.com/questions/10132005/startswith-method-in-java-returns-true-on-an-empty-string-how) agrees with me)

If this is what you wanted to do, to make the launcher useless, then go right ahead. Otherwise, you probably shouldn't merge this

EDIT: Confirmed with testing, it does. Nice job.

divadsn:
Seriously? Can you just calm the duck down?

divadsn:
Duplicate of #902

PR #904, opened 2017-11-08
dynamic status bar gradient #fixderps

saikrishna279:
This is like a switch to turn ON/OFF the annoying white gradient over the status bar area when a dark/black wallpaper is applied. Especially noticeable on LED Displays
It is replaced by a fully transparent background when a dark/black wall is applied
The gradient is shown when a light/bright wallpaper is applied

-added the algorithm
-fixed derp for lwp 
-removed "Show top shadow" from settings.

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/Launcher.java
+++ b/app/src/main/java/ch/deletescape/lawnchair/Launcher.java
@@ -28,6 +28,7 @@ import android.app.ActivityOptions;
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.app.SearchManager;
+import android.app.WallpaperManager;
 import android.appwidget.AppWidgetHostView;
 import android.appwidget.AppWidgetManager;
 import android.appwidget.AppWidgetProviderInfo;
@@ -46,8 +47,11 @@ import android.content.res.Configuration;
 import android.database.sqlite.SQLiteDatabase;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
+import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
+import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.Drawable;
 import android.os.AsyncTask;
 import android.os.Bundle;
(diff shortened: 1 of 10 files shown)


suphon-t:
I think [this](https://github.com/Deletescape-Media/Lawnchair/blob/alpha/app/src/main/java/ch/deletescape/lawnchair/dynamicui/ExtractedColors.java#L222) would be more efficient since the color extraction service does not run on every resume and is already used in other parts.

Also, what about live wallpapers?

saikrishna279:
The gradient is set when a lwp is applied without the concern to the lwp's brightness. And also, this runs a Lumination (light intensity) calculation algorithm, not color extraction algorithm

suphon-t:
But it's not efficient as it is run on every resume and the method I suggested already does this.

saikrishna279:
I am afraid to say that I was referring to the status bar gradient and not the Light Status bar

suphon-t:
My point is that you can still use that value to show/hide gradient.

saikrishna279:
Well, you can calculate the color of the wallpaper, but, you cannot calculate the brightness of the wallpaper.

I mean, the color can be red, blue, green or whatever, but, you cannot get the brightness of each pixel and calculate the overall brightness of the wallpaper

divadsn (commented):
It's an overkill for battery life to calculate live in background, especially for lwp with more than 20 updates per second (=20 FPS).

saikrishna279:
It only runs once the activity is restarted and also runs in separate thread and has no impact on battery life.

AND I already mentioned that, if the wallpaper that has been set is a live wallpaper, it doesn't calculate, instead, it sets the gradient and leaves it that way.

Please go through the code once for reference

nyancrimew:
As @paphonb already mentioned in the Color extraction service which already exists there is actually indeed a brightness calculation because this is needed for the Dock color. Please add your calculations to that service so they only run once.

suphon-t:
It is actually [calculated on every resume](https://i.imgur.com/TsPgYCx.png), not on restart. And what if the user want to have a live wallpaper without the shadow?

saikrishna279:
@paphonb we can implement a switch to decide wether the user needs shadow on the lwp. The brightness is calculated on every resume ( atleast that is the method i thought ) on activity resume, the brightness is calculated as the wallpaper may have been changed. Else, we should implement a method to keep  a watch to tell if the wallpaper is set (which is a broadcast imo) and we need a broadcast listener. Calculating brightness on every resume is better than using a broadcast listener imo. We can improve the mechanism anyway

suphon-t:
A switch would make this feature completely pointless and calculating on every resume is worse. A typical user would go back to the home screen more often than changing wallpapers.

divadsn:
To be re-opened if there is a real need.

PR #922, opened 2017-11-15
Add Lawnfeed popup for left swipe gesture

divadsn:


suphon-t (approved):

nyancrimew (changes requested):
The webpage to download Lawnfeed needs to be available before this is merged (atleast in it's minimal form)

divadsn:
@deletescape done, it's dynamic so it gets the latest download link from my S3 server.

PR #924, opened 2017-11-16
Don't show up Lawnchair in app search results

divadsn:


TheComputerGuy96:
Lawnchair-ception:
Lawnchair-ception

divadsn:
@TheComputerGuy96 yes, this fix will prevent that.

PR #928, opened 2017-11-17
Find a better way to restart Lawnchair

divadsn:


divadsn:
@deletescape do it.

nyancrimew:
Moved to #929

PR #946, opened 2017-11-20
Exit overview mode when clicking on actions

enricocid:
Show the home when going back from settings, widgets or wallpaper picker


divadsn:
Thanks, we'll merge it when we will work on shortcuts.

PR #955, opened 2017-11-21
Utilities: Improve how we detect package state

msfjarvis:
* Add isPackageInstalled and isAvailablePackage helpers
  to detect installed and enabled packages respectively
  make the blacklisted applications filter method
  utilise it.

This is untested for obvious reasons but should work fine.


divadsn:
We do have those methods for isPackageAvailable in the PackageManagerHelper afaik, maybe you could add them there :wink:

msfjarvis:
Sure will fix in a few.

PR #956, opened 2017-11-21
Utilities: Cleanup

msfjarvis:
* Change method visibilities wherever recommended.
* Inline some variables.
* Some more changes.

Moot point: Should the getSystemProperty method stop reflecting
and be reused to use a shell with getprop?

Signed-off-by: Harsh Shandilya <msfjarvis@gmail.com>


divadsn (approved):
Looks good to me, let's see what Codacy says :P

msfjarvis:
Works properly so figured why not

divadsn:
Shet, build failed.

PR #964, opened 2017-11-22
Detect pirate app store

burakdede0:
- Detects Aptoide which has pirated apps in it.


msfjarvis:
Has my -1. Aptoide isn't necessarily a piracy hub and blocking it is stupid and hugely counter productive and will affect more legitimate users than pirates.

burakdede0:
If you think so

PR #1033, opened 2017-12-15
Answering the call of duty

Ej-leone:


nyancrimew:
What are you exactly trying to do here?

Frontear:
Looks like they tried to update translation.

PR #1118, opened 2018-01-18
Double tap to sleep and wake

fosiczka:
Please add the functions of a double tap to wake and especially double tap to sleep. We need this, thanks 👍


divadsn:
This PR doesn't add anything new, hence closing.

Issue #1119, opened 2018-01-19
Add possibility to configure dock opacity


remiberthoz:
### Description
In situations where the background colour of the dock is extracted from the wallpaper, and the latter is light, the dock itself is almost transparent. (See attached screenshot!)

### Expected Behavior
Maybe an opacity slider in the Dock settings?

### May I try this as my first PR?
I know a bit of Android programming, I'd actually love to try this myself, if you don't mind.

### Screenshot

![screenshot_20180119-172442](https://user-images.githubusercontent.com/1943662/35160712-43c9f83e-fd3e-11e7-96b2-5a39d6d2dc57.png)

PR #1120, opened 2018-01-19
Add possibility to configure dock opacity

remiberthoz:
It's now possible to toggle a manual dock opacity option. The manual configuration is a slider ranging from 0% to 100%. If manual is off, automatic is used and unchanged from previous versions.

#1119

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/dynamicui/ExtractedColors.java
+++ b/app/src/main/java/ch/deletescape/lawnchair/dynamicui/ExtractedColors.java
@@ -206,6 +206,11 @@ public class ExtractedColors {
                 hotseatColor = ColorUtils.setAlphaComponent(FeatureFlags.INSTANCE.useDarkTheme(FeatureFlags.DARK_ALLAPPS) ? Color.BLACK : Color.WHITE, (int) (0.25f * 255));
             }
         }
+        boolean useCustomOpacity = Utilities.getPrefs(context).getHotseatShouldUseCustomOpacity();
+        if (useCustomOpacity) {
+            float customOpacity = Utilities.getPrefs(context).getHotseatCustomOpacity();
+            hotseatColor = ColorUtils.setAlphaComponent(hotseatColor, (int) (customOpacity * 255));
+        }
         return hotseatColor;
     }
 
--- a/app/src/main/java/ch/deletescape/lawnchair/preferences/IPreferenceProvider.kt
+++ b/app/src/main/java/ch/deletescape/lawnchair/preferences/IPreferenceProvider.kt
@@ -129,6 +129,8 @@ interface IPreferenceProvider {
     val iconPackPackage: String
     val hotseatIconScale: Float
     val hotseatHeightScale: Float
+    val hotseatShouldUseCustomOpacity: Boolean
+    val hotseatCustomOpacity: Float
     val enablePhysics: Boolean
     val snowflakeSizeScale: Float
     val snowflakesNum: String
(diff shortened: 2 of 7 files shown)


nyancrimew:
Looks good to me codewise, in case nobody has time earlier I will test it on monday. Until then I would really welcome it if you could stash your commits.

remiberthoz:
I'm not super familiar with git yet, but I assume you meant squash and not stash?

nyancrimew:
@TiiXel Yep, shouldn't reply to Pull Requests while I'm almost falling asleep I guess ;)

divadsn:
I'd rather make it with a slider only instead of having an additional switch and keep the default value, just like we did with other customization options.

LinuxLlama:
Is there a reset option?

remiberthoz:
@deletescape How should I handle this? In my understanding if I squash locally I won't be able to push. Do I need to open a new pull request on a new branch?

> I'd rather make it with a slider only instead of having an additional switch and keep the default value, just like we did with other customization options.

That's what I had in mind, too. Only the default value is not a fixed one, it depends on the extracted colour.

divadsn:
@TiiXel the color can be switched off and can be independent of the opacity.

@Treemindr there is no such thing as reset entirely :P

nyancrimew:
@TiiXel You can simply force push to your branch, it will update the Pull Request

remiberthoz:
@divadsn If I'm not mistaken, in current versions, the dock opacity is chosen based on the wallpaper color, regardless of the *Dock uses wallpaper colors* setting. If that is correct, adding only a slider in the settings would not allow users to use either automatic, or manual dock opacity. Users might want to use the extracted color, with or without a custom opacity ; others might want to use a dark/light themed dock, with or without custom opacity.

divadsn:
@TiiXel That's not the point, first of all the color extraction switch does work, as I am having my dock in white color and not in the extracted one from the wallpaper. And second, the opacity slider would be enough, there is no need for another switch for custom opacity.

remiberthoz:
@divadsn I know it works, no problem with that. My point is, if I remove the custom dock opacity toggle, the app behaviour will change or uses cases will be missing.

Currently, users *not* using the extracted color have their dock opacity set to either 12%, 18% or 25% depending on the chosen theme.

If I understand correctly, what you suggest is to link the opacity slider to the *Use extracted color* toggle.

Users might like to:
* Not use extracted color, use default opacity (currently possible)
* Not use extracted color, set custom opacity (will miss if I remove the toggle)
* Use extracted color, use extracted opacity (currently possible) (will miss if I remove the toggle)
* Use extracted color, set custom opaticy

It's possible I do not understand what you meant, sorry.

------------------------------

I realize the *Transparent dock* setting is redundant with the suggested slider and toggle, as one might set the dock being not transparent but with no opacity. I think it should be removed.

divadsn:
After looking up some files and commits from the past, it looks good, will recheck it later and merge it.

remiberthoz:
Thanks. Tell me if it needs any modifications!

In my last comment, I dumped:

> I realize the Transparent dock setting is redundant with the suggested slider and toggle, as one might set the dock being not transparent but with no opacity. I think it should be removed.

I thought about it and realized I have no idea how to maintain users configuration in case I removed this setting. If anyone can help with this, before you merge!

divadsn:
No problem, I will add some migration helper functions and then it will be merged :)

remiberthoz:
Thanks!

remiberthoz:
Closes #1119

Issue #1104, opened 2018-01-01
Lawnfeed Update notification - EVEN with lawnfeed not installed


shaderecker:
Lawnchair tells me to update my lawnfeed installation, even when I haven't installed it at all.

## Description

I use lawnchair as my launcher and everytime when the app is freshly started, e.g. after device boot, when the launcher is killed / not in memory anymore, there pops up a notification to update my lawnfeed installation. I haven't installed any version of lawnfeed, so this shouldn't pop up. It also happens on launcher rotation.

## Expected Behavior

There should be no popup, if no version of lawnfeed is installed.

## Actual Behavior

The notification pops up regardless of lawnfeed installed or not.

## Steps to Reproduce

1. Reboot device / kill launcher / rotate launcher
2. Popup will appear

## Environment

* Device: Samsung Galaxy S3
* Android version: 7.1.2
* Launcher version: 1.1.0.1742
* Rom: lineageOS-20171214

## Screenshot
![screenshot_20171230-230521](https://user-images.githubusercontent.com/16366556/34470699-c6b0bac2-ef37-11e7-9d69-1ea9406581db.png)

divadsn:
@TheHADILP I don't believe you, there is no chance the popup can come up if you haven't Lawnfeed installed, as confirmed by @fonix232 multiple times in other issues.

shaderecker:
I can assure you, I don't have lawnfeed installed. Neither now nor have I had it in the past.

fonix232:
I have double checked the code responsible for this popup, and there's no way this happens if the Lawnfeed package (or any other package with the same package name) doesn't exist on the device. 

The checks are [here](https://github.com/LawnchairLauncher/Lawnchair/blob/6422bb7d1da43eaef548a5c2a8522a81b9a4ce09/app/src/main/java/ch/deletescape/lawnchair/Utilities.java#L1094), first statement disables Lawnfeed integration on CI (non-Play Store) builds, the second one runs a few tests:

- Checks if the Google app is enabled, if not, won't load Lawnfeed
- Checks if the Lawnfeed app (or any other app with the same package name since we don't do signature checks) is installed and enabled
- Checks if the Lawnfeed app's version is higher than 1655

[...]

PR #1124, opened 2018-01-21
Fix "Lawnfeed outdated" popup in #1104

divadsn:


divadsn:
What why? This fixes the popup when Google app is not installed or Lawnfeed not running.

PR #1157, opened 2018-02-28
F-Droid build flavor (WIP)

divadsn:
Re-opens #806

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -93,6 +93,7 @@ android {
         }
         fdroid {
             dimension "store"
+            buildConfigField "boolean", "AWARENESS_API_ENABLED", "false"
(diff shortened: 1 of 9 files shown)


divadsn:
To be upstreamed and updated with other build flavors.

divadsn:
I think we are good for now, a fdroid flavor is already present in v2, so will close that.

ildar:
Where?
https://github.com/LawnchairLauncher/Lawnchair/search?q=fdroid&unscoped_q=fdroid

divadsn:
@ildar v2 sources aren't public yet

ildar:
ah! ok, will try to follow up then.

PR #1159, opened 2018-03-01
Properly handle storage permission on 8.1+

nyancrimew:
Prompt the user for the Storage permission with an explaination and disable blur without it to prevent crashes.


divadsn:
Travis, why tf?

PR #1172, opened 2018-03-20
Fix Popup related bugs

nyancrimew:
If you want this to look halfway decent just enable force RTL in dev options... Please just don't touch anything if you have built this, nothing really works and everything could result in a crash.

I should've probably just watched anime instead....


divadsn:
@deletescape No anime for you!

PR #1179, opened 2018-03-24
Apply hidden apps filter on item select

msfjarvis:
Rip out the apply entry from the overflow menu and instead
call the app filter method everytime an item is selected or
unselected for slightly better UX.

Signed-off-by: Harsh Shandilya <harsh@prjkt.io>


msfjarvis:
I did a bulk cleanup rather than just removing the strings orphaned by this change, let me know if you guys want me to reset the commit and only include the relevant cleanup.

PR #1186, opened 2018-03-26
PopupContainer used the wrong inflater (Fix #1178)

### LINKED ISSUE #1178: Long press on icons with dark theme no text visible
##Description
When the dark or grey theme is activated and you long press any icon on the home screen and in the app drawer, you get a black box without text for selection. Everything is fine in the light theme.

## Expected Behavior
There should be white text on black background

## Actual Behavior
There is a black box without text (i am guessing the text is there but in black too)

## Steps to Reproduce
Long Select any icon on home screen or app drawer , till the selection window opens

## Environment
Newest lawnchair 1.1.0.1868
On M-Horse pure 2 with stock firmware
Previous version of lawnchair was fine

nyancrimew:


divadsn:
screenshot_1522087409

divadsn (approved):
See my attached screenshot for reference.

PR #1187, opened 2018-03-26
Don't override custom icons on abort (Fix #1185)

### LINKED ISSUE #1185: Don't override custom icons if no new one is selected
Ideally, if you don't pick a new icon, the last one should be kept rather than the default one. 

Original title: app with a customized icon from icon packs will be restored to the default icon if you enter the icon selection interface but don't pick a new one(hit back button directly, for example)

nyancrimew:


divadsn (approved):
mergin the sheet

PR #1232, opened 2018-05-09
Always assume that an app could be on the sd card, fixes  #507

### LINKED ISSUE #507: Moved apps to sdcard wont show after reboot
Many apps i use i have transferred them to external storage. After phone reboot i loose all these apps from desktops or app bar drawer. Even after phone finish loading all transferred apps, Lawnchair won't restore theme to their previous locations. Other Launchers do keep in memory these apps and that don't happen. Original pixel launcher do have this bug. Here some pics of Sony's launcher after a reboot: icons for transferred apps look different but after loading finish, they revert to normal. Can u make smth similar?

_Sent from my Sony F3211 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Hocuri:
@fonix232 According to the commit history, you are the one who submitted the original code. Do you know if there could be any drawbacks if this method just always returns true? According to the method comment, false-positives can already happen (it "doesn't guarantee that the app is on the SD card")

### DIFF
--- a/app/src/main/java/ch/deletescape/lawnchair/util/PackageManagerHelper.java
+++ b/app/src/main/java/ch/deletescape/lawnchair/util/PackageManagerHelper.java
@@ -38,9 +38,10 @@ public class PackageManagerHelper {
     /**
      * Returns true if the app can possibly be on the SDCard. This is just a workaround and doesn't
      * guarantee that the app is on SD card.
+     * Currently, this always returns true as a previous approach did not work properly.
      */
     public static boolean isAppOnSdcard(PackageManager pm, String packageName) {
-        return isAppEnabled(pm, packageName, PackageManager.GET_UNINSTALLED_PACKAGES);
+        return true;
     }
 
     public static boolean isAppEnabled(PackageManager pm, String packageName, int flags) {


fonix232:
I only committed the refactor of the code, that's why my name's there. TBH I have no idea how this will affect other systems.

shubhchan:
@fonix232 This bot is fast and we can do this too

Hocuri:
@fonix232 Do you know who was the original author of the code was?

Apart from that, on my device it worked and as I already pointed out, false-positives could already happen before, so I am quite optimistic that nothing will be broken (but of course other proposals are welcome).

fonix232:
@Hocceruser Honestly, no idea. Might be @deletescape since pretty much before the refactor, the code was either his, or Google's. You'd need to dig into the git history for a better answer I'm afraid.

Hocuri:
OK, what happens after my change is: If I
* add an icon of App A to Lawnchair's home screen
* force close Lawnchair (this could maybe also happen if the device runs out of ram)
* uninstall A
then the icon of A will remain on the home screen, greyed out, until the user deletes it manually. 

But I do not think that this scenario is very likely, the issue of icons disappearing after a restart seems more important to me. Still, other proposals are welcome as my PR is not a perfect solution.

Anyway, this is probably what the original programmer wanted to avoid.

Hocuri:
Now, will you accept my pull request or not?

divadsn:
@Hocceruser We had no time to discuss it internally

PR #1234, opened 2018-05-11
ptBR update

caiorrs:


divadsn:
@caiorrs Translations are done over at Crowdin.

PR #1240, opened 2018-05-13
Fix backup ui to be consistent with the rest of settings

suphon-t:


divadsn (approved):
Works fine, merging it! :)

PR #1263, opened 2018-06-12
Feedly Api

ghost:
Can you put rss feed (for example feedly window) instead of google feed on the left screen using feedly api.


divadsn:
What the heck?!

Issue #1266, opened 2018-06-12
Launcher Crashes on Date Click


addiittya2006:
## Description
A crash occurs in the Launcher when there's no calender app available on Date Click on the home screen. {and date can have accidental clicks}

## Expected Behavior
It should just state that there is no calendar app available.

## Actual Behavior
log:
```
06-13 00:53:14.977 31423-31423/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: ch.deletescape.lawnchair.plah, PID: 31423
    android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=content://com.android.calendar/time/1528831394952 bnds=[744,163][1029,274] }
        at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2127)
        at android.app.Instrumentation.execStartActivity(Instrumentation.java:1801)
        at android.app.Activity.startActivityForResult(Activity.java:4495)
        at android.app.Activity.startActivity(Activity.java:4810)
        at ch.deletescape.lawnchair.pixelify.DateWidgetView.onClick(Unknown Source:60)
        at android.view.View.performClick(View.java:6303)
        at android.view.View$PerformClick.run(View.java:24828)
        at android.os.Handler.handleCallback(Handler.java:789)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6798)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

## Steps to Reproduce
![date_app_launcher_crash](https://user-images.githubusercontent.com/7330961/41312736-7d4ed650-6ea5-11e8-88ed-22a4fbf7f438.gif)

## Environment
* Device: OnePlus 3T
* Android version: 8.0.0
* Launcher version: 1.2.0.1878
* Rom: Oxygen OS 5.0.3

PR #1267, opened 2018-06-12
Fixed Issue #1266

gathurian:
Fixed issue #1266 where the launcher would crash if there was no available calendar app. Now, a dialog pops up, which informs the user about the fact and prevents the launcher from crashing.


nyancrimew:
Remove unused imports and this PR will be accepted

divadsn:
No activity since 2 weeks.

PR #1287, opened 2018-06-29
Add shortcut blacklist

divadsn:


divadsn:
Someone could maybe help out with a less fucked up way to use the MultiSelectRecyclerViewAdapter without having hardcoded mess.

divadsn:
Works, to be reviewed.

PR #1288, opened 2018-06-29
Revert edge glow effect and restore damped overscroll

suphon-t:
App drawer overscroll is already backported so this should be too.


app/src/main/java/ch/deletescape/lawnchair/PagedView.java
```diff
@@ -31,6 +31,7 @@
 import android.os.Bundle;
 import android.util.AttributeSet;
 import android.util.DisplayMetrics;
+import android.util.Log;
```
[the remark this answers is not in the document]
divadsn: Fuck

divadsn:
@paphonb a toggle in the Settings would be nice to have, otherwise works fine.

app/src/main/java/ch/deletescape/lawnchair/PagedView.java
```diff
@@ -1803,6 +1804,10 @@ protected void onScrollInteractionBegin() {
     }
 
     protected void onScrollInteractionEnd() {
+        if (mShowLawnfeedPopup) {
+            mShowLawnfeedPopup = false;
+            Utilities.showLawnfeedPopup(getContext());
```
divadsn, line 1809: Is working, but how do we bring back the old fade effect?

app/src/main/java/ch/deletescape/lawnchair/PagedView.java
```diff
@@ -1425,6 +1386,29 @@ public int getScrollForPage(int index) {
...
     }
 
+    // This curve determines how the effect of scrolling over the limits of the page dimishes
+    // as the user pulls further and further from the bounds
+    private float overScrollInfluenceCurve(float f) {
+        f -= 1.0f;
+        return f * f * f + 1.0f;
+    }
+
+    protected float acceleratedOverFactor(float amount) {
```
divadsn, line 1396: We also might wanna check if the device is in battery save mode to disable spring animation.

divadsn (changes requested):
Left some comments, otherwise nothing to complain.

app/src/main/java/ch/deletescape/lawnchair/PagedView.java
suphon-t: Actually this is just damping the scroll distance, not as demanding as the spring animation.

PR #1314, opened 2018-07-19
add new strings.xml pt-rBR

ojuara1987:
Can I help with the Brazilian Portuguese translation?


divadsn:
@ojuara1987 Please head over to Crowdin for translation

ojuara1987:
@divadsn I already have the strings.xml pt-rBR translated as I publish? thank you

fonix232:
@ojuara1987 We only use Crowdin for translations. Importing from other places would mean we'd have to import all the stuff back to Crowdin instead of using the already designed flow.

ojuara1987:
@ fonix232 ok thanks, but how do I use the Crowdin need to install on the computer? I really wanted to help with the translation.

bohrabhijeet:
@ojuara1987 You just need to make an account on Crowdin and go to translate.lawnchair.info in order to translate in your preferred language. You can access it from PC as well as mobile.

ojuara1987:
@abhijeetbohra Ok I'll do it right now, thank you

PR #1384, opened 2018-10-14
Typo: Unescaped '

dakshin-k:
Two translations, ca and fr had an apostrophe not preceded with \ which caused a build error.


zhai:
please delete my question, operate error

PR #1395, opened 2018-11-13
Fix section sorting issue on some devices with Simplified Chinese locale

junyuecao:


app/src/main/java/ch/deletescape/lawnchair/allapps/AlphabeticalAppsList.java
```diff
@@ -311,7 +311,9 @@ private void onAppsUpdated() {
         // coalesce sections
         Locale curLocale = mLauncher.getResources().getConfiguration().locale;
         TreeMap<String, ArrayList<AppInfo>> sectionMap;
-        boolean localeRequiresSectionSorting = curLocale.equals(Locale.SIMPLIFIED_CHINESE);
+        boolean localeRequiresSectionSorting =
+                Locale.SIMPLIFIED_CHINESE.getLanguage().equals(curLocale.getLanguage())
```
nyancrimew, line 315: This equality should already be granted with the general equality check over Locale itself from what I can see. Could you please show an example of behavior before and after this patch and what the expected behavior is?

junyuecao: On some devices I think they are different and `equals` return `false`.

fonix232: I think in this case you should compare baseLocale instead, no?

junyuecao: BaseLocale and getBaseLocale() are not public.

junyuecao: Any update or better solution?

PR #1465, opened 2019-04-08
Fix multiline labels in folders

neekless:
Attempt to fix multiline labels in folders being cut off. Use HomeLabelRows since that's what folders use.

Please take a look and see if it works as I'm not pro enough. :/

### DIFF
--- a/src/com/android/launcher3/DeviceProfile.java
+++ b/src/com/android/launcher3/DeviceProfile.java
@@ -484,6 +484,7 @@ public class DeviceProfile implements LawnchairPreferences.OnPreferenceChangeLis
     }
 
     private void updateFolderCellSize(float scale, DisplayMetrics dm, Resources res) {
+        LawnchairPreferences prefs = Utilities.getLawnchairPrefs(mContext);
         int folderLabelRowCount = prefs.getHomeLabelRows();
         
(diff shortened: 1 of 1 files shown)


nyancrimew:
LGTM, will cherry-pick and test later today

src/com/android/launcher3/DeviceProfile.java
```diff
@@ -484,11 +484,13 @@ private void updateAvailableFolderCellDimensions(DisplayMetrics dm, Resources re
     }
 
     private void updateFolderCellSize(float scale, DisplayMetrics dm, Resources res) {
+        int folderLabelRowCount = prefs.getHomeLabelRows();
```
nyancrimew: There was an issue here that `prefs` wasn't a member in this class yet, to get rid of all our instantiations of this all over the class I fixed this and applied your commit internally.

neekless: Oh right. I just followed the format in the workspace and all apps sections, I should've known I missed it out.
The fix was to add `LawnchairPreferences prefs = Utilities.getLawnchairPrefs(mContext);` correct?

nyancrimew: Yup that would have fixed it!

PR #1483, opened 2019-04-09
Fix multiline labels not truncating when >2 lines

neekless:
For singleline labels, if labels are too long it is truncated with an ellipsis.
For multiline labels, it is coded to be max 2 lines, but labels that are too long to fit in 2 lines instead becomes 3 lines instead of being truncated. This results in the labels getting cut off (depending on grid size I think).
This is an attempt to fix this.

@deletescape Once again I'm not sure if this will work. I based this on https://github.com/LawnchairLauncher/Lawnchair/pull/744 in V1 where 2-line labels are truncated. I hope you can test it. :)


nyancrimew:
This doesn't seem to actually fix the issue and doesn't visually change anything in my quick tests

neekless:
Okay I will close this

PR #1488, opened 2019-04-09
SystemShortcut: Add uninstall option

Jertlok:
This allows to easily uninstall application on the popup
activity triggered by long pressing on the applications.

### DIFF
--- a/src/com/android/launcher3/popup/PopupDataProvider.java
+++ b/src/com/android/launcher3/popup/PopupDataProvider.java
@@ -78,7 +78,8 @@ public class PopupDataProvider implements NotificationListener.NotificationsChan
                 new SesameSettings(),
                 new SystemShortcut.AppInfo(),
                 new SystemShortcut.Widgets(),
-                new SystemShortcut.Install()
+                new SystemShortcut.Install(),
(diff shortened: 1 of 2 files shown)


Jertlok:
@deletescape, let me know if I shouldn't write this in Java - nevertheless this is worth pushing to DirtyUnicorns Launcher3, so I did not waste my time.

nyancrimew:
We generally do accept Java, and for smaller changes like these I think I wouldn't bother with rewriting in Java either.

divadsn:
I leave that to other team members, as I know that in the past we were rather not liking that.

Edit: "[...] so far we said we kinda don't wanna clutter the menu too much but idk"

Jertlok:
I do understand that, we might move that under the custom edit action, but that wouldn't make too much sense at that point you could juust press on the (i) and uninstall the application with the same amounts of actions :+1:

divadsn:
@Jertlok Or drag the app and move to the "Uninstall" at the top of the screen 😉

Jertlok:
Force pushed as I have noticed that I added the shortcut with the maximum priority, fixed that just in case you will reconsider this shortcut :+1:

PR #1489, opened 2019-04-09
CustomBottomSheet: App label improvements

Jertlok:
* Move from EditText to ExtendedEditText from L3 with a small
  method added to force disabling the AutoFill avaiability.
  Also set the maximum app label size to 30 as it sounds reasonable
  to do so, and disable suggestions for label inputs.

### DIFF
--- a/src/com/android/launcher3/ExtendedEditText.java
+++ b/src/com/android/launcher3/ExtendedEditText.java
@@ -34,6 +34,7 @@ public class ExtendedEditText extends EditText {
 
     private boolean mShowImeAfterFirstLayout;
     private boolean mForceDisableSuggestions = false;
+    private boolean mForceDisableAutoFill = false;
 
     /**
      * Implemented by listeners of the back key.
@@ -140,5 +141,19 @@ public class ExtendedEditText extends EditText {
             }
         }
         hideKeyboard();
+    }
+
+    /**
+     * This method allows to force AutoFill availability.
+     *
+     * @param forceDisableAutoFill whether to disable AutoFill or not.
+     */
+    public void setAutoFill(boolean forceDisableAutoFill) {
+        mForceDisableAutoFill = forceDisableAutoFill;
+    }
+
+    @Override
+    public int getAutofillType() {
+        return mForceDisableAutoFill ? AUTOFILL_TYPE_NONE : super.getAutofillType();
     }
 }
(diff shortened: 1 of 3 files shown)


Jertlok:
I am up for restoring the keyboard suggestions or eventually increase the label size.

divadsn:
Haven't ever seen any password manager asking to autofill, but I often use keyboard suggestions.

Jertlok:
@divadsn, if you long press on the app label text box a "Fill" option appears.
Ensure you haven't edited the app before.

divadsn:
This solution might be better instead:
https://stackoverflow.com/questions/45731372/disabling-android-o-auto-fill-service-for-an-application

Jertlok:
> This solution might be better instead:
> https://stackoverflow.com/questions/45731372/disabling-android-o-auto-fill-service-for-an-application

Unfortunately those clean solutions did not work for me, I wish that AutoFill XML tag (or the programmatic way for every view) worked with Google AutoFill.

divadsn:
@Jertlok Google SmartLock isn't playing the same afaik, but I will take a look today myself.

ghost:
>   Also set the maximum app label size to 30 as it sounds reasonable

May I suggest 32 instead? It's a more computer-generated number. 32bits, 32GB, etc

msfjarvis:
> > Also set the maximum app label size to 30 as it sounds reasonable
> 
> May I suggest 32 instead? It's a more computer-generated number. 32bits, 32GB, etc

That's really not how this works...

suphon-t:
I can’t remember, but what does ExtendedEditText offer start over a normal EditText?

src/com/android/launcher3/ExtendedEditText.java
```diff
@@ -141,4 +142,18 @@ public void reset() {
         }
         hideKeyboard();
     }
+
+    /**
+     * This method allows to force AutoFill availability.
+     *
+     * @param forceDisableAutoFill whether to disable AutoFill or not.
+     */
+    public void setAutoFill(boolean forceDisableAutoFill) {
```
zeletrik, line 151: I know I have no right to judge the code. But in the coding aspect, it's really misleading. The `setAutoFill` should accept the autofill enable value, not the disable value.
You may call it `disableAutoFill` but then it's a negation so `enableAutoFill` is much nicer in that way too.
 In the `getAutofillType` method just switch the ternary operator. Then the comment can be removed where it's used _(CustomBottomSheet#108)_ as well.

RikKoedoot:

PR #1529, opened 2019-04-24
Extend Firefox search provider to work with fennec_fdroid

brianclemens:

PR #1550, opened 2019-05-09
Add support for Fennec F-Droid as search provider

xerz-one:
[Fennec F-Droid](https://f-droid.org/en/packages/org.mozilla.fennec_fdroid/) is a fork of Firefox maintained by the F-Droid community, "_focused on removing any proprietary bits found in official Mozilla's builds_". It is mostly the same browser as the latest stable version of Firefox except for branding.

P.S.: Maybe the user should be allowed to select Firefox version instead of doing `firstOrNull`?


xerz-one:
Whoops, just noticed https://github.com/LawnchairLauncher/Lawnchair/pull/1529! Not sure to close the PR tho, since the other one set the program ID at the last position, making the program prefer to use Firefox Beta or Aurora over Fennec.

PR #1605, opened 2019-06-22
Add gesture action that opens recents using the accessibility service

tonyk69:
Also added default and greek translation stings for the action.


tonyk69:
if anything needs to be changed please inform me of it

PR #1728, opened 2019-08-31
AbstractStateChangeTouchController: Do NULL check in updateProgress()

aviraxp:


aviraxp:
Somehow I get NULL pointer here when pressing HOME. Not sure if it is a aosp bug or...

PR #1767, opened 2019-10-06
Make 3rd party apps restore backups automatically.

oddlyspaced:
This pull request would let 3rd party apps restore lawnchair backups with an intent call, directly.


RikKoedoot:

PR #1808, opened 2019-11-11
Fix adaptive icon shapes for legacy shortcuts

Shingyx:
Fixed by updating `createIconBitmap` to wrap icon drawables using `AdaptiveIconCompat`.

Below are screenshots of a Pixel 2 emulator, configured for square icons. The "Sleep" shortcut is a legacy shortcut using an adaptive icon, and the two shortcuts to the right of it are legacy shortcuts which don't use adaptive icons.

Before | After
:-------------------------:|:-------------------------:
bad | good

I _think_ this is the right place for this change but I could be wrong. Please let me know what you think.

Thanks!

### DIFF
--- a/src/com/android/launcher3/graphics/LauncherIcons.java
+++ b/src/com/android/launcher3/graphics/LauncherIcons.java
@@ -146,9 +146,13 @@ public class LauncherIcons implements AutoCloseable {
             Resources resources = mPm.getResourcesForApplication(iconRes.packageName);
             if (resources != null) {
                 final int id = resources.getIdentifier(iconRes.resourceName, null, null);
+                Drawable drawable = resources.getDrawableForDensity(id, mFillResIconDpi);
+                if (drawable != null) {
+                    drawable = AdaptiveIconCompat.wrap(drawable);
+                }
                 // do not stamp old legacy shortcuts as the app may have already forgotten about it
                 return createBadgedIconBitmap(
-                        resources.getDrawableForDensity(id, mFillResIconDpi),
+                        drawable,
                         Process.myUserHandle() /* only available on primary user */,
                         0 /* do not apply legacy treatment */);
             }


paphonb:
This still doesn't fix the full drawable that's shown while dragging. Maybe try wrapping [here](https://github.com/LawnchairLauncher/Lawnchair/blob/5a8a4be9c6ecbbbcffaf813c32cc60dc2bf46833/src/com/android/launcher3/shortcuts/DeepShortcutManager.java#L196) instead?

suphon-t:
This still doesn't fix the full drawable that's shown while dragging. Maybe try wrapping [here](https://github.com/LawnchairLauncher/Lawnchair/blob/5a8a4be9c6ecbbbcffaf813c32cc60dc2bf46833/src/com/android/launcher3/shortcuts/DeepShortcutManager.java#L196) instead?

Shingyx:
image

Dragging seems to be fine, unless you mean something else?

I don't think `DeepShortcutManager` ever gets involved when dealing with legacy shortcuts. It seems to be all handled inside `InstallShortcutReceiver`. From what I could see anyway.

1. https://github.com/LawnchairLauncher/Lawnchair/blob/5a8a4be9c6ecbbbcffaf813c32cc60dc2bf46833/src/com/android/launcher3/InstallShortcutReceiver.java#L246-L249
2. https://github.com/LawnchairLauncher/Lawnchair/blob/5a8a4be9c6ecbbbcffaf813c32cc60dc2bf46833/src/com/android/launcher3/InstallShortcutReceiver.java#L518-L519
3. https://github.com/LawnchairLauncher/Lawnchair/blob/5a8a4be9c6ecbbbcffaf813c32cc60dc2bf46833/src/com/android/launcher3/InstallShortcutReceiver.java#L652-L656

suphon-t:
Ah my bad, thought you mean the new shortcuts with legacy icons. Will take a look at this when I can.

Shingyx:
Rebased onto the latest alpha branch.

Not sure if this is worth mentioning, but I've actually been running my own fork of Lawnchair with this fix ever since I made this PR, and saw no issues.

Missheathermarie:
Were watching it but no need for us to jump yet

baby

PR #1851, opened 2019-12-03
Apps in the drawer

thor9898:
I think there must be a switch to be able to see all the apps in the drawer although there are some in the suggestion or in the tabs.


suphon-t:
If you want to request for a new feature, create an issue and not a pull request.

thor9898:
Sorry.

PR #1853, opened 2019-12-03
Folders

thor9898:
As I came from Action launcher there are som features I loved and I would love to see in this launcher.
It would be cool if we can add the apps to the folder from an option from the options displayed when you click on the 3 dots on the apps.
Also it would be cool if we have an option to order the apps their names.
And last but not least it woul be amazing we could have different options to open folders on cover mode, as I see more efficient to open them swiping down than up, but it's always nice to have different options :)
Thank you very much.


suphon-t:
Duplicate of #1852

PR #1933, opened 2020-02-03
Update Edge support

TheLBall:
Microsoft recently changed the activity of Edge search. 

This code is updated, but now only works on newer versions of Edge.


RikKoedoot:

PR #2087, opened 2021-02-26
Infinite scroll support

Dav357:
Re-up of  #2055 (made a mess with git...)
Same code changes, less translations files (they were all the same anyway)


RikKoedoot:

PR #2097, opened 2021-03-22
[11-dev] Fix icon size and label size reload, implement launcher restart

AgentFabulous:


ghost (approved):

Pandemic-XDA:
Icon pack changer works not so good like on lc for a10
Example: when you download glass icon pack by ambor here
https://play.google.com/store/apps/details?id=com.ambor.iconpack
It changes on previous lawnchair but not on the dev 11 build.

PR #2099, opened 2021-03-23
[11-dev] Lawnchair: Swipe down on Workspace to expand statusbar

AgentFabulous:


ghost (approved):

suphon-t:
https://user-images.githubusercontent.com/8080853/112642481-efc64b80-8e75-11eb-9316-82b95a288211.mp4


I noticed that this sometimes trigger when swiping through workspace pages. Would it be better to implement this as a [TouchController](https://github.com/LawnchairLauncher/Lawnchair/blob/a1f1fc2e057f560ac2f360d3b4ac549859ead33b/src/com/android/launcher3/util/TouchController.java) like other gestures? That way the DragLayer would handle priorities for you. 

Launcher3 already comes with [this](https://github.com/LawnchairLauncher/Lawnchair/blob/b9fb3b0d0f9b345b1ae26951499134fd1607a11e/quickstep/src/com/android/launcher3/uioverrides/touchcontrollers/StatusBarTouchController.java) status bar gesture that uses quickstep. Perhaps we could modify it to use StatusBarManager when quickstep isn't available.

AgentFabulous:
>  2564-03-26.20.57.15.mp4 
> I noticed that this sometimes trigger when swiping through workspace pages. Would it be better to implement this as a [TouchController](https://github.com/LawnchairLauncher/Lawnchair/blob/a1f1fc2e057f560ac2f360d3b4ac549859ead33b/src/com/android/launcher3/util/TouchController.java) like other gestures? That way the DragLayer would handle priorities for you.
> 
> Launcher3 already comes with [this](https://github.com/LawnchairLauncher/Lawnchair/blob/b9fb3b0d0f9b345b1ae26951499134fd1607a11e/quickstep/src/com/android/launcher3/uioverrides/touchcontrollers/StatusBarTouchController.java) status bar gesture that uses quickstep. Perhaps we could modify it to use StatusBarManager when quickstep isn't available.

I don't think the default statusbar gesture would work in our case. That requires a signature level permission, which then starts TouchInteractionService and populates SystemUIProxy. I'll look into TouchController.

suphon-t:
> I don't think the default statusbar gesture would work in our case.

That's why I suggest modifying it to use StatusBarManager when the proxy is not available.

PR #2127, opened 2021-04-11
Fix OneUI Gesture crash

DanGLVK:


divadsn:
Any description? Testing?

DanGLVK:
> Any description? Testing?

Was tested by an individual who was trying another fixed version, i built the app from source with the fix and sent it to him to confirm it's working 

https://user-images.githubusercontent.com/32078304/114310772-57110c00-9ac2-11eb-8166-086d699b3857.mp4

suphon-t:
This is weird. The method you added isn't referenced anywhere at all. How does it fix this?

DanGLVK:
> This is weird. The method you added isn't referenced anywhere at all. How does it fix this?

Fact is, while Lawnchair isn't calling this it seems One UI is, having a blank class seems to be enough to satisfy One UI and not cause it to crash Lawnchair

suphon-t:
Can I look at the crash log once again? On the version without this fix.

suphon-t:
I actually misinterpreted your changes. Can you verify that it still works when compiling with `optimized` variant? I'm worried proguard might strip the method out.

DanGLVK:
> I actually misinterpreted your changes. Can you verify that it still works when compiling with `optimized` variant? I'm worried proguard might strip the method out.

Decompiling the optimized build seems to show the method  is still there.

suphon-t (approved):

PR #2191, opened 2021-07-27
Fix android studio compilation problem

qian-jiahong:


divadsn:
Jesus no, first: Reflections do not work on Android.

And second: https://github.com/LawnchairLauncher/lawnchair/blob/11-dev/README.md#contribute-code

PR #2193, opened 2021-08-03
Home screen grid : increase max size

AirOne70:
Allow users to use a bigger grid. This is very useful on big screens or when icons size are reduced.


suphon-t:
Allowing up to 15 makes it hard to select a value. Maybe reduce it to around 10?

AirOne70:
12 steps may be too much yes. The old version of Lawnchair was allowing up to 20 rows / columns.
Limiting to 10 may not be enought depending on your DPI. I'm using 10 rows already on my personal setup.
On the other side, is a 3 rows setup really useful?

lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
```diff
@@ -69,13 +69,13 @@ fun HomeScreenPreferences() {
                 label = stringResource(id = R.string.home_screen_columns),
                 adapter = prefs.workspaceColumns.getAdapter(),
                 step = 1f,
-                valueRange = 3.0F..7.0F
+                valueRange = 3.0F..15.0F
```
ghost:
```suggestion
                valueRange = 3.0F..10.0F
```

lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
```diff
@@ -69,13 +69,13 @@ fun HomeScreenPreferences() {
...
             )
             SliderPreference(
                 label = stringResource(id = R.string.home_screen_rows),
                 adapter = prefs.workspaceRows.getAdapter(),
                 step = 1f,
-                valueRange = 3.0F..7.0F,
+                valueRange = 3.0F..15.0F,
```
ghost:
```suggestion
                valueRange = 3.0F..10.0F,
```

ghost (changes requested):
As per @paphonb’s comment, I would suggest a maximum value of 10.

AirOne70 (commented):
Changed to 10 as requested. :)

ghost:
Cheers!

RikKoedoot:
The appdrawer columns are not going up to 10, also if we are going with high numbers, is a slider still the best option?

RikKoedoot:
Also dock icons can also only be changed up to 7

PR #2199, opened 2021-08-05
Add option to disable labels on home screen

AirOne70:
Not sure if it's the best way to do it, but it's working as expected.

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
@@ -101,6 +101,7 @@ fun HomeScreenPreferences() {
             SwitchPreference(
                 prefs.showHomeLabels.getAdapter(),
                 label = stringResource(id = R.string.show_home_labels),
+                showDivider = showHomeLabels
             )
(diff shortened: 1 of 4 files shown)


lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
```diff
@@ -95,14 +97,25 @@ fun HomeScreenPreferences() {
...
-                label = stringResource(id = R.string.label_size),
-                adapter = prefs.textSizeFactor.getAdapter(),
-                step = 0.1f,
-                valueRange = 0.5F..1.5F,
-                showAsPercentage = true,
-                showDivider = false
+            val showHomeLabels by prefs.showHomeLabels.observeAsState()
+            SwitchPreference(
+                prefs.showHomeLabels.getAdapter(),
+                label = stringResource(id = R.string.show_home_labels),
```
suphon-t, line 103: Should hide the divider when it's toggled off

suphon-t (changes requested):
Other than the requested change, this looks fine

AirOne70:
Good catch, done!

PR #2208, opened 2021-08-19
Fix spring position out of range

HuixingWong:
… cannot be less than the min value


paphonb:
I think it'd be better to set `mRange` as the minimum of `range` and `mDelta`. That way we would have to bound the range only once.

suphon-t (changes requested):
I think it'd be better to set `mRange` to the minimum of `range` and `mDelta`. That way we would have to bound the range only once.

HuixingWong:
> I think it'd be better to set `mRange` to the minimum of `range` and `mDelta`. That way we would have to bound the range only once.

Indeed, thanks!

PR #2263, opened 2021-10-25
change TextColorSecondary token to a generated value

kshatriya-abhay:
* fixes white on white color in Work profile tab strip (light mode)

before:
image

after:
image


ghost (approved):

suphon-t:
I think it'd be better to keep it `#de000000` for consistency. Can you try fixing `StaticColorToken` instead?

PR #2349, opened 2021-11-30
TODO implemented as b/77875376 is fixed

IMPranshu:


suphon-t:
No you can't just remove that. It wouldn't even compile.

IMPranshu:
Ok Sorry for that let me dig around some more and find another way to implement it

PR #2360, opened 2021-12-04
Distinguish between Simplified Chinese and Traditional Chinese

### LINKED ISSUE #2207: Simplified Chinese is incorrectly displayed
The language set on my phone is simplified Chinese, but the language it displays is traditional Chinese

### LINKED ISSUE #2335: Simplified Chinese not correctly recognized
When the system language is simplified Chinese, Lawnchair launcher did not recognize the language correctly.

Instead of showing simplified Chinese, the launcher showed traditional Chinese.

JohnsonRan:
To solve #2207 and #2335
I found that Chinese translations are only  one folder named `values-zh`,but Simplified and Traditional Chinese is not the same folder name,like `values-zh-rCN`,`values-zh-rTW`,one is Simplified Chinese, another is Traditional Chinese,like these folders:
Screenshot_20211205-174035_Chrome~2
So i renamed and created new folder,hope it can fix the issue


ghost:
Hello! We use Crowdin for localization, so the files and directories for localization generally shouldnʼt be modified directly. This issue was caused by a Crowdin setting not being correctly configured. As of commit 9879b59ff1c7b56f981c1e6da6a5baf5e8321637, the problem has been resolved. Lawnchair is translated into both traditional and simplified Chinese on Crowdin, but you are more than welcome to suggest changes to the existing translations: https://lawnchair.crowdin.com/lawnchair. Thank you!

PR #2392, opened 2021-12-27
Allow icon pack dashboards to change icon pack externally

DarkionAvey:
Hello!

This is a port from legacy Lawnchair, but with a new intent name to match the new package name of the app '`app.lawnchair`'

Check out the video below to see how it looks, or download the test icon pack (attached) to test it yourself 

How to call:
```
        val lawnchair = Intent("app.lawnchair.APPLY_ICONS", null)
        lawnchair.putExtra("packageName", packageName)
        startActivity(lawnchair)
```
[(test icon pack) app-release.zip](https://github.com/LawnchairLauncher/lawnchair/files/7780210/app-release.zip)

Tested on Android 12, but I assume it works fine on older versions because the UI is made in Compose

https://user-images.githubusercontent.com/11511782/147472892-193dd5ff-4490-4301-8c0e-bf62573741dc.mp4

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
@@ -62,36 +62,35 @@ class ApplyIconPackActivity : AppCompatActivity() {
                     CompositionLocalProvider(LocalPreferenceInteractor provides viewModel<PreferenceViewModel>()) {
                         val iconPacks by LocalPreferenceInteractor.current.iconPacks.collectAsState()
                         val isIconPackValid = iconPacks.any { it.packageName == packPackageName }
-                        Box(modifier = Modifier.fillMaxSize()) {
-                            Surface(
-                                shape = RoundedCornerShape(topStart = 24.dp, topEnd = 24.dp),
-                                color = MaterialTheme.colorScheme.surface,
-                                modifier = Modifier
-                                    .fillMaxWidth()
-                                    .align(Alignment.BottomCenter)
-                            ) {
-                                Box(
-                                    Modifier
+                        if (iconPacks.isNotEmpty()) {
+                            Box(modifier = Modifier.fillMaxSize()) {
+                                Surface(
+                                    shape = RoundedCornerShape(topStart = 24.dp, topEnd = 24.dp),
+                                    color = MaterialTheme.colorScheme.surface,
+                                    modifier = Modifier
                                         .fillMaxWidth()
-                                        .navigationBarsOrDisplayCutoutPadding()
+                                        .align(Alignment.BottomCenter)
                                 ) {
-                                    Column(
-                                        modifier = Modifier
+                                    Box(
+                                        Modifier
                                             .fillMaxWidth()
-                                            .padding(25.dp)
-                                            .verticalScroll(rememberScrollState())
+                                            .navigationBarsOrDisplayCutoutPadding()
                                     ) {
-                                        if (isIconPackValid && packPackageName != null) {
-                                            ApplyIconPackShowcase(
-                                                iconPackPackageName = packPackageName,
-                                                iconPackName = packName,
-                                                iconPackIcon = packIcon
-                                            )
-                                        } else {
-                                            InvalidIconPackPlaceholder(
-                                                iconPackName = packName,
-                                                iconPackIcon = packIcon
-                                            )
+                                        Column(
+                                            modifier = Modifier
+                                                .fillMaxWidth()
+                                                .padding(25.dp)
+                                                .verticalScroll(rememberScrollState())
+                                        ) {
+                                            if (isIconPackValid && packPackageName != null) {
+                                                ApplyIconPackShowcase(
+                                                    iconPackPackageName = packPackageName,
+                                                    iconPackName = packName,
+                                                    iconPackIcon = packIcon
+                                                )
+                                            } else {
+                                                finish()
+                                            }
                                         }
                                     }
                                 }
@@ -122,30 +121,6 @@ class ApplyIconPackActivity : AppCompatActivity() {
                 )
             }
             HorizontalSpacer(height = 8.dp)
-        }
-    }
-
-    @Composable
-    private fun ColumnScope.InvalidIconPackPlaceholder(
-        iconPackName: String?,
-        iconPackIcon: Bitmap?
-    ) {
-        ShowcaseIcon(bitmap = iconPackIcon)
-        Text(
-            text = stringResource(R.string.invalid_icon_pack_description).format(
-                Locale.getDefault(),
-                iconPackName
-            ),
-            modifier = Modifier.fillMaxWidth(),
-            textAlign = TextAlign.Center,
-            color = MaterialTheme.colorScheme.onSurface
-        )
-        HorizontalSpacer()
-        FullWidthButton(
-            muted = true,
-            text = stringResource(id = android.R.string.cancel)
-        ) {
-            finish()
         }
     }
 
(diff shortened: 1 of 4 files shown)


lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+import android.os.Bundle
+import androidx.activity.compose.setContent
+import androidx.appcompat.app.AppCompatActivity
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.Surface
```
suphon-t: Please use material3 Surface instead

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+
+        setContent {
+            LawnchairTheme {
+                ProvideWindowInsets {
+                    CompositionLocalProvider(LocalPreferenceInteractor provides viewModel<PreferenceViewModel>()) {
+                        val iconPacks by LocalPreferenceInteractor.current.iconPacks.collectAsState()
+                        val isIconPackValid = iconPacks.any { it.packageName == packPackageName }
+                        Box(modifier = Modifier.fillMaxSize()) {
+                            Surface(
+                                shape = RoundedCornerShape(topStart = 25.dp, topEnd = 25.dp),
```
suphon-t: Should be 24dp

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+                                            .verticalScroll(rememberScrollState())
+                                    ) {
+                                        if (isIconPackValid && packPackageName != null) {
+                                            ApplyIconPackShowcase(
+                                                iconPackPackageName = packPackageName,
+                                                iconPackName = packName,
+                                                iconPackIcon = packIcon
+                                            )
+                                        } else {
+                                            InvalidIconPackPlaceholder(
```
suphon-t: I think we should just finish the activity in this case

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+            Surface(
+                shape = CircleShape,
+                modifier = Modifier
+                    .height(80.dp)
+                    .aspectRatio(1f)
+                    .align(Alignment.CenterHorizontally),
+                color = MaterialTheme.colorScheme.surface
+            ) {
+                Image(
+                    painter = rememberImagePainter(bitmap),
```
suphon-t: I think we can do `bitmap = bitmap.asImageBitmap()` instead of using Coil

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }
+    }
+
+    @OptIn(ExperimentalCoilApi::class)
+    @Composable
+    private fun ColumnScope.CircularIcon(bitmap: Bitmap?) {
+        if (bitmap != null)
+            Surface(
+                shape = CircleShape,
```
suphon-t: Let's keep the original icon shape instead of cropping to a circle

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+            text = iconPackName,
+            modifier = Modifier.fillMaxWidth(),
+            fontWeight = FontWeight.Medium,
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
+            fontSize = 20.sp,
+            textAlign = TextAlign.Center,
+            color = MaterialTheme.colorScheme.onSurface
+        )
+        HorizontalSpacer(height = 8.dp)
```
suphon-t: I think 16 might look better

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+
+    @Composable
+    private fun ColumnScope.ApplyIconPackShowcase(
+        iconPackPackageName: String,
+        iconPackName: String,
+        iconPackIcon: Bitmap?
+    ) {
+        val hapticFeedback = LocalHapticFeedback.current
+        var iconPacksPreference by preferenceManager().iconPackPackage.getAdapter()
+        CircularIcon(bitmap = iconPackIcon)
```
suphon-t: Add some space above the icon

lawnchair/src/app/lawnchair/ui/preferences/ApplyIconPackActivity.kt
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Composable
+    private fun ColumnScope.ApplyIconPackShowcase(
+        iconPackPackageName: String,
+        iconPackName: String,
+        iconPackIcon: Bitmap?
+    ) {
+        val hapticFeedback = LocalHapticFeedback.current
+        var iconPacksPreference by preferenceManager().iconPackPackage.getAdapter()
```
suphon-t: You don't need the adapter here. The pref entry has a `set` method.

suphon-t (changes requested):

DarkionAvey:
I implemented all of your comments except for finishing the activity when the icon pack is invalid. I don't know how to achieve that while re-composition is triggered. The only thing I can think of is to implement a quick icon pack validation by checking the package's intents against `app.lawnchair.ui.preferences.PreferenceViewModelKt#iconPackIntents` and short-circuiting if the icon pack is invalid, all before `setContent`. What do you suggest? 
Thanks

HLA55:
###

suphon-t:
> I implemented all of your comments except for finishing the activity when the icon pack is invalid. I don't know how to achieve that while re-composition is triggered. The only thing I can think of is to implement a quick icon pack validation by checking the package's intents against `app.lawnchair.ui.preferences.PreferenceViewModelKt#iconPackIntents` and short-circuiting if the icon pack is invalid, all before `setContent`. What do you suggest?
> Thanks

I think you can just call `finish()` in a side effect block, since the Activity would still be in scope

DarkionAvey:
> > I implemented all of your comments except for finishing the activity when the icon pack is invalid. I don't know how to achieve that while re-composition is triggered. The only thing I can think of is to implement a quick icon pack validation by checking the package's intents against `app.lawnchair.ui.preferences.PreferenceViewModelKt#iconPackIntents` and short-circuiting if the icon pack is invalid, all before `setContent`. What do you suggest?
> > Thanks
> 
> I think you can just call `finish()` in a side effect block, since the Activity would still be in scope

So I experimented with different ways, and I found that checking `iconPacks.isNotEmpty()` is the simplest one. 
Let me know what you think. 

Also, there's a little visual disturbance because `backgroundDimEnabled` is set to true so the window will dim then go back to normal after few seconds if the icon pack is invalid. I think it's OK since the user won't end up seeing this anyway (i.e., only happens with invalid packs)

DarkionAvey:
Any updates on whether this will be implemented or not?

ghost:
Hello, @DarkionAvey! Thanks for your interest in contributing, and sorry for the wait. We’d be happy to merge this pull request with some changes, but because you appear to have deleted the Lawnchair fork on which this PR is based, no changes can be suggested or made. Would you be able to create a new Lawnchair fork, copy the changes from this pull request to it, and open a new pull request? I’ll then be able to suggest some changes. Thanks in advance, and sorry once more for the delay.

Shine19958:
Pull

PR #2400, opened 2022-01-01
Add an option to hide the "Turn off work apps" button

zehuanli:
The "Turn off work apps" sometimes overlaps with the app icons when there are lots of apps installed in the work profile. From my own experience, it can be accidentally clicked, which suddenly terminates all the running work apps and causes data loss.

A preference entry has been added to the "App Drawer" -> "General" tab to allow the user to choose whether the "Turn off work apps" button should be displayed in the work profile tab in the app drawer.

image

image


zehuanli:
nvm

PR #2427, opened 2022-01-19
Disable Force Dark on the launcher

DanGLVK:
In order to avoid potentially miscolored icons when users have this option active on their devices


suphon-t:
Can you move this to [`LauncherTheme`](https://github.com/LawnchairLauncher/lawnchair/blob/12-dev/res/values/styles.xml#L31) instead? As other themes doesn't extend `AppTheme`.

DanGLVK:
> Can you move this to [`LauncherTheme`](https://github.com/LawnchairLauncher/lawnchair/blob/12-dev/res/values/styles.xml#L31) instead? As other themes doesn't extend `AppTheme`.

Is this alright?

PR #2463, opened 2022-02-07
Add "Show Component Name" feature

evowizz:
This new preference would allow developers and designers who want to contribute to https://github.com/LawnchairLauncher/lawnicons or any other icon pack to easily obtain the component name (package name and activity name) of the target app if needed.

By default, the flag is set to false, but it can be enabled in the Debug menu. Once enabled, the component name of any app can be seen under the Customize dialog (accessible from both the app drawer and the home screen) thanks to the `description` and `showDescription` parameters of the `PreferenceGroup` component. Here's how it looks like:

feature/show-component-name


ghost (approved):

ghost:
Thanks, @evowizz!

evowizz:
Well that was fast! Thanks! 🎉

Issue #2541, opened 2022-04-17
[SUGGESTION] More icon shapes from Lawnchair 2


severtheskyline:
### Describe the bug

As of Lawnchair 12 Alpha 5, Lawnchair only have 4 icon shapes:
-The system one
-Circular
-Squircle
-Rounded square 
Now, this is better than any other vanilla launcher out there, but is nowhere near as close as the old Lawnchair 2 (And Lawnchair 10, before the rewrite).
There are the next missing shapes:
-Square
-One UI
-Cilinder
-Teardrop

### Steps to reproduce

1. Go to 'General'
2. Click on 'Icon Shapes'
3. See the only icon shapes there. 

### Expected behavior

More icon shapes (Or at least parity between the old Lawnchair and the slick new Lawnchair). 

### Screenshots

Old:
![Screenshot_20220417-144244](https://user-images.githubusercontent.com/79670265/163731404-6b373e2f-6092-43ed-9495-452c51a1d065.png)
New:
![Screenshot_20220417-144250](https://user-images.githubusercontent.com/79670265/163731417-57f88902-6695-42b8-971c-f71ac2c8b6ab.png)

### Device information

- Device: [ LG Stylo 6 ]
- OS: [ LG UX 9 on Android 10 ]
- App version [ 12.0.0 Alpha 5 ]

### Additional context

I would love the old squared ones to be in the new Lawnchair, they look similar to Android Go, and pretty good.

yasanglass:
Lawnchair 12 supports 8 icon shapes and based on my brief testing, they all work and are unique. Not sure why the UI only lets the user choose between 4 options. I can do a simple PR that adds support for this but there is probably a reason behind the limitation.

https://github.com/LawnchairLauncher/lawnchair/blob/c868e32c97d9630f93f10ffd143c3fd736cca6d9/lawnchair/src/app/lawnchair/icons/shape/IconShape.kt#L284-L302

severtheskyline:
It could be a UI bug, but why select such specific shapes to appear? Seems interesting.

RikKoedoot:
I asked @paphonb, he said the code is copied from V2, but he does not remember why we did not use all shapes.

SuperDragonXD:
Someone has made a PR for this issue now. I guess we should just wait until they merge it.

yasanglass:
PR was merged. Probably should close this now. @PandoraHayami

severtheskyline:
Right. Lovely!

PR #2544, opened 2022-04-21
Add missing icon shapes to the preferences GUI

yasanglass:
There are only 3 icon shape options (excluding *System*) in the GUI but internally Lawnchair 12 supports 8 icon shapes. Based on my testing they all work and are unique. This PR adds them all to the preferences GUI.

Old/Current icon shapes list:
photo_2022-04-21_12-30-14

New icon shapes list:
photo_2022-04-21_12-54-55

On the previous GUI, `Cupertino` is named *Rounded Square*. Since I wanted to add all of the options and `RoundedSquare` is another shape, `Cupertino` is now named *Cupertino* in the GUI. All the options in the GUI match the shape names in the code now except `Sammy` which is shown as *One UI* in the preferences GUI since based on #2541 it looks like that should be the name the users see.

### DIFF
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -151,8 +151,13 @@
     <string name="icon_shape_label">Icon Shape</string>
     <string name="icon_shape_system">System</string>
     <string name="icon_shape_circle">Circle</string>
+    <string name="icon_shape_square">Square</string>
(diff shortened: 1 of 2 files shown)


ghost (approved):

ghost:
Thank you, @DariaHamrahPaytakht and @yasandev!

PR #2637, opened 2022-05-23
Persian Calendar on Glance

yasanglass:
Adds Persian Calendar support to Glance + A system to manually reload the date content on Glance as it was not previously possible.

The changes also allow other contributors to add their own local calendar system to the project easily.

The Persian Date is always shown in Farsi for now but I will create an additional PR to make the Persian Date respect the language of the device later on as the library used for the Persian Date does not support this feature (yet).

### DIFF
--- a/AndroidManifest-common.xml
+++ b/AndroidManifest-common.xml
@@ -78,7 +78,8 @@
         android:label="@string/derived_app_name"
         android:largeHeap="@bool/config_largeHeap"
         android:restoreAnyVersion="true"
-        android:supportsRtl="true" >
+        android:supportsRtl="true"
+        tools:replace="android:label">
 
         <!-- Intent received when a session is committed -->
         <receiver
--- a/lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
+++ b/lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
@@ -8,11 +8,14 @@ import android.icu.text.DateFormat
 import android.icu.text.DisplayContext
 import android.os.SystemClock
 import android.util.AttributeSet
+import app.lawnchair.preferences.getAdapter
 import app.lawnchair.preferences2.PreferenceManager2
 import app.lawnchair.preferences2.subscribeBlocking
 import app.lawnchair.smartspace.model.SmartspaceCalendar
 import app.lawnchair.util.viewAttachedScope
 import com.android.launcher3.R
+import com.patrykmichalik.preferencemanager.first
+import com.patrykmichalik.preferencemanager.firstBlocking
 import saman.zamani.persiandate.PersianDate
 import saman.zamani.persiandate.PersianDateFormat
 import java.util.*
@@ -87,8 +90,14 @@ class IcuDateTextView @JvmOverloads constructor(
     override fun onFinishInflate() {
         super.onFinishInflate()
         preferenceManager2 = PreferenceManager2.getInstance(context)
-        preferenceManager2.smartspaceCalendar.subscribeBlocking(scope = viewAttachedScope) {
-            calendar = it
+        val calendarSelectionEnabled =
+            preferenceManager2.enableSmartspaceCalendarSelection.firstBlocking()
+        if (calendarSelectionEnabled) {
+            preferenceManager2.smartspaceCalendar.subscribeBlocking(scope = viewAttachedScope) {
+                calendar = it
+            }
+        } else {
+            calendar = preferenceManager2.smartspaceCalendar.defaultValue
         }
     }
 
--- a/lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
+++ b/lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
@@ -8,14 +8,23 @@ import android.icu.text.DateFormat
 import android.icu.text.DisplayContext
 import android.os.SystemClock
 import android.util.AttributeSet
+import app.lawnchair.preferences2.PreferenceManager2
+import app.lawnchair.preferences2.subscribeBlocking
+import app.lawnchair.smartspace.model.SmartspaceCalendar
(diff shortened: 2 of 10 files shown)


yasanglass:
Sorry for the force pushes, just a few small adjustments.

yasanglass:
- Improved how `IcuDateTextView.kt` listens to calendar changes.
- Smartspace custom calendar is now experimental & the default value is used when the experimental feature is disabled.
- Default calendar system for Farsi (Iran) is set to Persian in resources.

@paphonb should be ok now, please check again.

suphon-t (approved):

lawnchair/src/app/lawnchair/smartspace/model/SmartspaceCalendar.kt
```diff
@@ -0,0 +1,33 @@
...
+open class SmartspaceCalendar(@StringRes val nameResourceId: Int) {
+
+    companion object {
+        
+        fun fromString(value: String): SmartspaceCalendar {
+            return when (value) {
+                "persian" -> Persian
+                else -> Gregorian
+            }
+        }
```
ghost: An expression body could be used here for conciseness.

```suggestion
        fun fromString(value: String): SmartspaceCalendar =  when (value) {
            "persian" -> Persian
            else -> Gregorian
        }
```

AndroidManifest-common.xml
```diff
@@ -78,7 +78,8 @@
         android:label="@string/derived_app_name"
         android:largeHeap="@bool/config_largeHeap"
         android:restoreAnyVersion="true"
-        android:supportsRtl="true" >
+        android:supportsRtl="true"
+        tools:replace="android:label">
```
ghost, line 82: Just to confirm, is the addition of `tools:replace="android:label"` intentional? I’m unfamiliar with this attribute.

lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
```diff
@@ -39,27 +51,62 @@ class IcuDateTextView @JvmOverloads constructor(
...
+    private fun getTimeTextPersian(updateFormatter: Boolean): String {
+        if (formatterPersian == null || updateFormatter) {
+            val formatter = PersianDateFormat(
+                context.getString(R.string.smartspace_icu_date_pattern_persian),
+                PersianDateFormat.PersianDateNumberCharacter.FARSI
+            )
+            this.formatterPersian = formatter
+        }
+        return formatterPersian!!.format(PersianDate(java.lang.Long.valueOf(System.currentTimeMillis())))
+    }
```
ghost: I’d suggest avoiding the use of not-null assertion here.

```suggestion
    private fun getTimeTextPersian(updateFormatter: Boolean): String {
        val formatter = formatterPersian.takeIf { updateFormatter.not() } ?: PersianDateFormat(
            context.getString(R.string.smartspace_icu_date_pattern_persian),
            PersianDateFormat.PersianDateNumberCharacter.FARSI,
        ).also { formatterPersian = it }
        return formatter.format(PersianDate(java.lang.Long.valueOf(System.currentTimeMillis())))
    }
```

lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
```diff
@@ -39,27 +51,62 @@ class IcuDateTextView @JvmOverloads constructor(
...
+    private fun getTimeTextGregorian(updateFormatter: Boolean): String {
+        if (formatterGregorian == null || updateFormatter) {
+            val formatter: DateFormat = DateFormat.getInstanceForSkeleton(
+                context.getString(R.string.smartspace_icu_date_pattern_gregorian),
+                Locale.getDefault()
+            )
+            this.formatterGregorian = formatter
+            formatter.setContext(DisplayContext.CAPITALIZATION_FOR_BEGINNING_OF_SENTENCE)
+        }
+        return formatterGregorian!!.format(java.lang.Long.valueOf(System.currentTimeMillis()))
```
ghost: I’d suggest avoiding the use of not-null assertion here.

```suggestion
        val formatter = formatterGregorian.takeIf { updateFormatter.not() }
            ?: DateFormat.getInstanceForSkeleton(
                context.getString(R.string.smartspace_icu_date_pattern_gregorian),
                Locale.getDefault()
            ).also { formatterGregorian = it }
        return formatter.format(java.lang.Long.valueOf(System.currentTimeMillis()))
```

ghost (changes requested):

AndroidManifest-common.xml
yasanglass: It is. It basically tells gradle to use the `android:label` value from this manifest file and not from a conflicting one from the Persian Date library. Basically, it won't build with that dependency if this line is not added here.

lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
yasanglass: This is the exact original code and I just tried to not modify it and keep the same format even for the Persian Date function. It's now fixed on the new force push.

yasanglass:
@patrykmichalik Should be ok now.

AndroidManifest-common.xml
ghost: Got it. Thanks!

lawnchair/src/app/lawnchair/smartspace/IcuDateTextView.kt
ghost: Ah, sorry, I didn’t notice that you’d used the preexisting code. If I had, I wouldn’t have requested this change, as it’s not entirely related to the pull request. I would’ve probably made it myself after the merge. But if you’ve already applied it, I don’t think it hurts to keep it. My apologies for the error.

ghost (approved):

PR #2651, opened 2022-05-25
Fix Date & Time pref group being blank in some scenarios

yasanglass:
On the previously merged #2650 there is a small bug that this PR fixes.

If the selected calendar is Persian & then the experimental calendar customization is turned off, the "Date & Time" preference group will be blank because the code on `SmartspaceDateAndTimePreferences()` only checks the active calendar and does not take the experimental calendar system state into consideration. 

Instead of manually fixing the issue just for Persian, I added a new value to `SmartspaceCalendar` so this limitation could be enabled for other calendar systems later on easily.


yasanglass:
The unused imports were removed automatically by my IDE and I removed that in the force push as it's not related to this PR.

PR #2663, opened 2022-05-29
Home Screen Lock

### LINKED ISSUE #2480: Feature Request: Lock Home/Drawer
To prevent icons from mismove.

yasanglass:
Adds the option to lock home screen content, when enabled the user is unable to create any modifications to the Home screen content. The code is almost identical to how it's done on `9-dev`. (Closes #2480)

There can be a drawer lock feature as well but currently, it looks like all the Drawer interactions are meant to customize the Home content and there is no drawer customization to be locked (outside preferences).

The ***Widgets*** & **Customize** options on the apps' menu are removed when the lock is enabled and this change affects apps both on Drawer & Home since even if you *customize* an app from Drawer, it'll be reflected on the Home screen & the widgets button simply adds a widget to the Home screen.

So I wonder maybe it would be better if I rename the preference & move it to the General preferences as it's not really limited to the Home screen?

~~I also wonder if the "Add New Apps to Home Screen" should be automatically disabled when the *lock* is enabled?~~
I added a commit that disables "Add New Apps to Home Screen" & hides the preference in the GUI when the home screen is locked.

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
@@ -50,10 +50,14 @@ fun HomeScreenPreferences() {
     val prefs2 = preferenceManager2()
     val scope = rememberCoroutineScope()
     PreferenceLayout(label = stringResource(id = R.string.home_screen_label)) {
+        val lockHomeScreenAdapter = prefs2.lockHomeScreen.getAdapter()
         PreferenceGroup(heading = stringResource(id = R.string.general_label)) {
+            val addIconToHomeAdapter = prefs.addIconToHome.getAdapter()
             SwitchPreference(
-                prefs.addIconToHome.getAdapter(),
+                checked = !lockHomeScreenAdapter.state.value && addIconToHomeAdapter.state.value,
+                onCheckedChange = addIconToHomeAdapter::onChange,
                 label = stringResource(id = R.string.auto_add_shortcuts_label),
+                enabled = lockHomeScreenAdapter.state.value.not(),
             )
             SwitchPreference(
                 prefs.wallpaperScrolling.getAdapter(),
@@ -90,7 +94,7 @@ fun HomeScreenPreferences() {
                 subtitle = stringResource(id = R.string.x_by_y, columns, rows),
             )
             SwitchPreference(
-                adapter = prefs2.lockHomeScreen.getAdapter(),
+                adapter = lockHomeScreenAdapter,
                 label = stringResource(id = R.string.home_screen_lock),
             )
         }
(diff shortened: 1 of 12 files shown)


RikKoedoot:
I think it would be nice as well to have features grayed out in settings instead of disappear, and if you tap them you'd get a toast that says that locked homescreen is enabled. As in the past I've had many users be confused why they couldn't edit their homescreen when they enabled the lock accidentally.

yasanglass:
> I think it would be nice as well to have features grayed out in settings instead of disappear, and if you tap them you'd get a toast that says that locked homescreen is enabled. As in the past I've had many users be confused why they couldn't edit their homescreen when they enabled the lock accidentally.

I agree, I didn't do it originally because I thought I can only send an `adapter`. Now the switch is disabled when home screen is locked & it shows that it's disabled as well.

lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
```diff
@@ -50,10 +50,14 @@ fun HomeScreenPreferences() {
...
     PreferenceLayout(label = stringResource(id = R.string.home_screen_label)) {
+        val lockHomeScreenAdapter = prefs2.lockHomeScreen.getAdapter()
         PreferenceGroup(heading = stringResource(id = R.string.general_label)) {
+            val addIconToHomeAdapter = prefs.addIconToHome.getAdapter()
             SwitchPreference(
-                prefs.addIconToHome.getAdapter(),
+                checked = if (lockHomeScreenAdapter.state.value) false else addIconToHomeAdapter.state.value,
+                onCheckedChange = addIconToHomeAdapter::onChange,
                 label = stringResource(id = R.string.auto_add_shortcuts_label),
+                enabled = lockHomeScreenAdapter.state.value.not()
```
ghost:
```suggestion
                enabled = lockHomeScreenAdapter.state.value.not(),
```

lawnchair/src/app/lawnchair/ui/preferences/HomeScreenPreferences.kt
```diff
@@ -50,10 +50,14 @@ fun HomeScreenPreferences() {
     val prefs2 = preferenceManager2()
     val scope = rememberCoroutineScope()
     PreferenceLayout(label = stringResource(id = R.string.home_screen_label)) {
+        val lockHomeScreenAdapter = prefs2.lockHomeScreen.getAdapter()
         PreferenceGroup(heading = stringResource(id = R.string.general_label)) {
+            val addIconToHomeAdapter = prefs.addIconToHome.getAdapter()
             SwitchPreference(
-                prefs.addIconToHome.getAdapter(),
+                checked = if (lockHomeScreenAdapter.state.value) false else addIconToHomeAdapter.state.value,
```
ghost:
```suggestion
                checked = lockHomeScreenAdapter.state.value.not() && addIconToHomeAdapter.state.value,
```

ghost (changes requested):
Hello! Thanks for your interest in contributing. Here are some minor suggestions.

yasanglass:
@patrykmichalik both requests are applied on the new commit.

ghost:
Cheers!

ghost (approved):

suphon-t (approved):

ghost:
Thanks, @yasandev!

PR #2664, opened 2022-05-29
More Accessible Home Screen Lock Toggle

### LINKED ISSUE #1711: [Feature request] Make the desktop lock feature more accessible
Currently, if you want to move widgets or icons, rename apps and folders, resize widgets, etc., you would need to go all the way into desktop settings to unlock it, which is 1 tap & hold (might be different depending what gesture you set the home screen pop-up to) and 3 taps.

I think a "Lock/Unlock desktop" button should be added to the home screen pop-up menu (furthermore, add a toggle to turn that button on and off because I know there are people who don't want such button in the home screen pop-up). Additionally, it would be nice to have it mappable to gestures too (e.g., tap and hold to lock/unlock desktop).

yasanglass:
Adds Home Screen lock toggle to the home screen's pop-up menu to make it more accessible (Closes #1711).

This is meant to be merged after #2663 & will be updated if #2663 is modified. Only c49b78c631327a3e8917d18da97f93ea82d05388 is new on top of the changes on #2663.


suphon-t:
I think it could be hidden in the unlocked state to keep it clean because it probably won't be used often.

yasanglass:
@paphonb I can just add a simple switch for it but I suppose I should make a preference group for Lock in that case so it looks cleaner.

yasanglass:
I added the new toggle below the lock switch but I believe the whole Home Screen preferences could use a bit of cleaning up.

yasanglass:
I updated the code to use the new `setBlocking()` function from @patrykmichalik's PreferenceManager v1.0.7 instead of the previous way as it's much better in every aspect.

A `toggleBlocking()` function would make it even cleaner.

ghost:
Hello! I believe @paphonb’s suggestion is for the button to be visible when the home screen is locked, and gone when the home screen is unlocked. But let’s wait for his input.

yasanglass:
Uh, a one-way toggle button wouldn't be that useful in my opinion, especially since this is meant for accessibility. Maybe I can add two separate switches? One toggles the ***lock*** button and the other one toggles the ***unlock*** button. But I guess then it would be too complicated idk.

lawnchair/res/values/strings.xml
```diff
@@ -44,6 +44,8 @@
     <string name="max_folder_rows">Max. Folder Rows</string>
     <string name="home_screen_grid">Home Screen Grid</string>
     <string name="home_screen_lock">Lock Home Screen</string>
+    <string name="home_screen_unlock">Unlock Home Screen</string>
+    <string name="home_screen_lock_toggle_from_home_popup">Toggle Lock Button On Pop-up</string>
```
ghost:
```suggestion
    <string name="home_screen_lock_toggle_from_home_popup">Show Lock Button in Pop-Up</string>
```

ghost (changes requested):
Just one small suggestion from my end.

lawnchair/res/values/strings.xml
yasanglass: Fixed on the new force push.

ghost (approved):

ghost:
Thanks, @yasandev and @DariaHamrahPaytakht!

Issue #2654, opened 2022-05-26
[FEATURE REQUEST] Add search engine presearch


muznyo:
Please add Presearch.com to a list of available search engines.

yasanglass:
~~I tried adding it to the search engines and the website search works, but I have not been able to set up the in-app search yet. Maybe a series of website-only search providers could be useful to provide more variety?!~~

Just implemented it fully. Will create a PR for it soon.

PR #2671, opened 2022-05-31
Wikipedia Search Provider

yasanglass:
Adds Wikipedia as an alternative search provider. 

I could not find a way to make it work without the new `className` variable and it should not affect objects that have a blank `className` but if you know how I can avoid creating this new variable please let me know. 

I also do have a working [Presearch](https://presearch.com/) support which was requested on #2654 but I thought it would be better to separate the two PRs.

~~The code is working but I just need the proper graphics for both to open the pull requests.~~


yasanglass:
Added the Wikipedia logo.

yasanglass:
Fixed `ic_wikipedia`'s location.

ghost (approved):

ghost:
Cheers, @yasandev!

PR #2694, opened 2022-06-08
Allow setting default accent color from config.xml

yasanglass:
Improves how the default accent color is loaded; Instead of the default value only depending on the Android version of the device, it now tries to load a string value from the config.xml file. If the string value is valid, it uses the value from config.xml and if it is invalid, the previously used Android version-based system is used.


yasanglass:
I set the default value to "" so the behavior won't change. Since the old method does not fully translate to this new method, I was not sure which new default value would be best.

ghost (changes requested):
Thanks for your interest in contributing! Please correct me if I’m wrong, but I believe that if a custom accent color were defined via `config_default_accent_color`, no option would be marked as selected in the Accent Color sheet, and if the user were to select a new accent color, they would have no way of returning to the one defined via `config_default_accent_color`. This presents a bit of a problem. I think a color picker where you can select any color is a prerequisite here.

ghost (commented):
Thanks for your interest in contributing! Please correct me if I’m wrong, but I believe that if a custom accent color were defined via `config_default_accent_color`, no option would be marked as selected in the Accent Color sheet, and if the user were to select a new accent color, they would have no way of returning to the one defined via `config_default_accent_color`. This presents a bit of a problem. I think a color picker where you can select any color is a prerequisite here.

ghost:
Apologies for the duplicate comment. It seems that GitHub wonʼt let me delete either of the two comments.

yasanglass:
My plan was to add a fully custom color picker after this PR but I hadn't thought of it that way! I guess I can update this PR or create a new one once I have added the custom color picker.

Although other than the custom options in the UI, I think I need to add a "restore defaults" button as well so users can revert to the actual default value.

yasanglass:
I will create another PR later with an advanced color picker option added.

ghost:
Sounds good. Cheers!

yasanglass:
Since the custom/advanced color picker is now merged, I believe I can re-open this PR. The only change I made to this branch is that instead of `config_default_accent_color` being blank, It's now set to "default" which is an invalid value and has the same result but it's better than the config value being blank.

ghost (approved):

ghost:
Cheers!

PR #2716, opened 2022-06-20
Fix Glance title's alignment when both layout & text are RTL

yasanglass:
#2700 was previously merged to fix #2698 but after more testing, I found that it would only fix the issue when the layout is RTL but the title text itself isn't RTL (and it would cause issues for other scenarios). 

This PR removes the unneeded text alignment-related code from `IcuDateTextView.kt` and improves how text alignment is handled on `BcSmartSpaceCard.kt`.

Sorry for the issues introduced on #2700, I am going to test the next PRs much more carefully.

PR #2726, opened 2022-06-21
Fix Glance throwing NullPointerException

yasanglass:
Updates isStringRTL() to handle potential null exceptions which did happen on my tests after #2716 was merged.

Although looking at the code, the bug is a bit strange since the string can never be empty on line 156! Somehow it sometimes is.


ghost (approved):

ghost:
Thanks, @yasandev and @DariaHamrahPaytakht!

PR #2774, opened 2022-07-18
Add enabled/disabled icons to Switch in SwitchPreference

evowizz:
## Description

This change aims at providing better accessibility to Lawnchair by adding icons over the Switch in SwitchPreference.

Here's how it looks like:
pref-switch-icons

Note: While this change may be considered by many as a "general change," it may also be considered as a new feature for some, which is why I checked the following option below.

## Type of change

- [x] General change (non-breaking change that fixes typos/grammar, or adds additional content that isn't a bug or a feature)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/components/SwitchPreference.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/components/SwitchPreference.kt
@@ -19,7 +19,12 @@ package app.lawnchair.ui.preferences.components
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.*
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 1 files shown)


RikKoedoot:
Will this always be enabled or only when someone is using an accessibility service?

evowizz:
It will always be enabled. :)

yasanglass (approved):

SuperDragonXD (approved):

yasanglass:
We've decided that this change will not be added, therefore this PR will not be merged, unfortunately. Thanks for all your contributions though.

evowizz:
Perfect, thank you!

PR #2795, opened 2022-07-24
Add copy & paste buttons to Hex Color Picker

yasanglass:
Adds copy & paste buttons to Hex Color Picker to allow users to handle Hex codes easier.

The reason that `androidx.compose.ui.platform.ClipboardManager` is not used is that we initially tried it but it sometimes would paste blank strings; as a result, we ended up using `android.content.ClipboardManager`.

Test APK: [Lawnchair 12.1.0 Dev (6d63c73).zip](https://github.com/LawnchairLauncher/lawnchair/files/9175404/Lawnchair.12.1.0.Dev.6d63c73.zip)

Preview:
Peek 2022-07-24 12-59

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -362,6 +362,7 @@ dependencies {
     implementation "androidx.compose.ui:ui-tooling:$compose_version"
     implementation "androidx.compose.ui:ui-text-google-fonts:$compose_version"
     implementation "androidx.compose.foundation:foundation:$compose_version"
+    implementation "androidx.compose.material:material-icons-extended:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
     implementation "androidx.compose.material3:material3:1.0.0-alpha14"
     implementation "androidx.compose.runtime:runtime-livedata:$compose_version"
--- a/lawnchair/res/drawable/ic_content_paste.xml
+++ b/lawnchair/res/drawable/ic_content_paste.xml
@@ -1,0 +1,5 @@
+<vector android:autoMirrored="true" android:height="24dp"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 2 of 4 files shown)


suphon-t:
Please use [`ClickableIcon`](https://github.com/LawnchairLauncher/lawnchair/blob/995f89f9eb3bd26e4566617fda95c1c88fa1e751/lawnchair/src/app/lawnchair/ui/preferences/components/ClickableIcon.kt#L40) instead of `IconButton`. And for the icons themselves, please use an `ImageVector` from [`Icons.Rounded`](https://developer.android.com/reference/kotlin/androidx/compose/material/icons/Icons.Rounded).

yasanglass:
`ContentCopy` and `ContentPaste` icons require `androidx.compose.material:material-icons-extended` dependency. Should I add it for these icons?

kahanzammi:
Isn't it possible to include a "Color Picker" - that allow user to pick any specific color from the wallpaper?

yasanglass:
> Isn't it possible to include a "Color Picker" - that allow user to pick any specific color from the wallpaper?

That is unrelated to this PR and you should make a new issue for it. Anyways, it is possible but it might be a bit too much for a color picker in a launcher.

suphon-t:
> `ContentCopy` and `ContentPaste` icons require `androidx.compose.material:material-icons-extended` dependency. Should I add it for these icons?

Can you try and see how much the size increase will be? But it should be fine since we have proguard on release builds anyway.

yasanglass:
> Can you try and see how much the size increase will be? But it should be fine since we have proguard on release builds anyway.

There is a note about the size increase on the official documentation:

> Note that only the most commonly used icons are provided by default. You can add a dependency on androidx.compose.material:material-icons-extended to access every icon, but note that due to the very large size of this dependency you should make sure to use R8 / ProGuard to remove unused icons from your application.

But yeah, as you said it won't affect optimized builds. Just wanted to double-check. I will do the requested changes and push now.

yasanglass:
Done. @paphonb

PR #2840, opened 2022-08-14
Add uninstall button and it's fixes

marshmello61:
Added feature from DerpLauncher (Launcher3)

### DIFF
--- a/src/com/android/launcher3/popup/SystemShortcut.java
+++ b/src/com/android/launcher3/popup/SystemShortcut.java
@@ -237,6 +237,29 @@ public abstract class SystemShortcut<T extends Context & ActivityContext> extend
         }
     }
 
+    public static final Factory<BaseDraggingActivity> UNINSTALL = (activity, itemInfo) ->
+            PackageManagerHelper.isSystemApp(activity,
+                 itemInfo.getTargetComponent().getPackageName())
+                    ? null : new UnInstall(activity, itemInfo);
+
+
+    public static class UnInstall extends SystemShortcut<BaseDraggingActivity> {
+
+        public UnInstall(BaseDraggingActivity target, ItemInfo itemInfo) {
+            super(R.drawable.ic_uninstall_no_shadow, R.string.uninstall_drop_target_label,
+                    target, itemInfo);
+        }
+
+        @Override
+        public void onClick(View view) {
+            String packageName = mItemInfo.getTargetComponent().getPackageName();
+            Intent intent = new PackageManagerHelper(
+                    view.getContext()).getUninstallIntent(packageName);
+            mTarget.startActivitySafely(view, intent, mItemInfo);
+            AbstractFloatingView.closeAllOpenViews(mTarget);
+        }
+    }
+
     public static <T extends Context & ActivityContext> void dismissTaskMenuView(T activity) {
         AbstractFloatingView.closeOpenViews(activity, true,
             AbstractFloatingView.TYPE_ALL & ~AbstractFloatingView.TYPE_REBIND_SAFE);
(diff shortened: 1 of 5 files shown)


src/com/android/launcher3/popup/SystemShortcut.java
```diff
@@ -237,6 +239,38 @@ public void onClick(View view) {
...
+        public UnInstall(BaseDraggingActivity target, ItemInfo itemInfo) {
+            super(R.drawable.ic_uninstall_no_shadow, R.string.uninstall_drop_target_label,
+                    target, itemInfo);
+        }
+
+        @Override
+        public void onClick(View view) {
+            try {
+                Intent intent = Intent.parseUri(view.getContext().getString(R.string.delete_package_intent), 0)
+                    .setData(Uri.fromParts("package", mItemInfo.getTargetComponent().getPackageName(),
```
yasanglass, line 264: `getTargetComponent` can return null and cause null pointer exception.

src/com/android/launcher3/util/PackageManagerHelper.java
```diff
@@ -256,21 +256,34 @@ public static IntentFilter getPackageFilter(String pkg, String... actions) {
...
 
+    public static boolean isSystemApp(Context context, String pkgName) {
+        return isSystemApp(context, null, pkgName);
+    }
+
     public static boolean isSystemApp(Context context, Intent intent) {
+        return isSystemApp(context, intent, null);
+    }
+
+    public static boolean isSystemApp(Context context, Intent intent, String pkgName) {
```
yasanglass: It would be better if there were only 2 functions:

- Remove `isSystemApp(Context context, String pkgName)`
- Change `isSystemApp(Context context, Intent intent, String pkgName)` to only take `Context` and `String` package name.
- Modify `isSystemApp(Context context, Intent intent)` so it converts `Intent` to `String` package name internally and just passes it to the *main* function.

(do not apply the suggestion automatically/on web, I could not select the whole block it should replace)

```suggestion
        public static boolean isSystemApp(Context context, Intent intent) {
        PackageManager pm = context.getPackageManager();
        // Get the package name for intent
        String packageName = null;
        if (intent != null) {
            ComponentName cn = intent.getComponent();
            if (cn == null) {
                ResolveInfo info = pm.resolveActivity(intent, PackageManager.MATCH_DEFAULT_ONLY);
                if ((info != null) && (info.activityInfo != null)) {
                    packageName = info.activityInfo.packageName;
                }
            } else {
                packageName = cn.getPackageName();
            }
        }
        return isSystemApp(context, packageName);
    }

    public static boolean isSystemApp(Context context, String packageName) {
        PackageManager pm = context.getPackageManager();
        // Check if the provided package is a system app.
        if (packageName != null) {
            try {
                PackageInfo info = pm.getPackageInfo(packageName, 0);
                return (info != null) && (info.applicationInfo != null) &&
                        ((info.applicationInfo.flags & ApplicationInfo.FLAG_SYSTEM) != 0);
            } catch (NameNotFoundException e) {
                return false;
            }
        } else {
            return false;
        }
    }
```

yasanglass (changes requested):
Thank you for the PR, looks good but a few small changes can make it better.

marshmello61:
@yasandev I am having time restrictions for quite some days. Can it be possible, you do changes accordingly and merge?

yasanglass:
> @yasandev I am having time restrictions for quite some days. Can it be possible, you do changes accordingly and merge?

Sure. I'll fix & merge it today myself.

yasanglass:
Thank you, @marshmello61.

I skipped the first requested change as it looks like similar codes were used on the same file already so it shouldn't be problematic.

The force push is because of a rebase update that I did before applying the new changes.

PR #2881, opened 2022-08-25
Reorganize parts of settings UI

SuperDragonXD:
## Description

This PR organizes the "General" preferences and adds additional strings/descriptions for "At a Glance" and "Home Screen" settings.

## Type of change

### DIFF
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -19,8 +19,6 @@
     <string name="iconPackPackageDefault" translatable="false">""</string>
     <string name="derived_app_name" tools:override="true" translatable="false">Lawnchair</string>
 
-    <!-- Common -->
-
     <!-- PreferenceDashboard -->
     <string name="settings">Settings</string>
     <string name="general_label">General</string>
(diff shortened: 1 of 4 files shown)


lawnchair/res/values/strings.xml
```diff
@@ -165,14 +163,16 @@
     <string name="wallpaper">Wallpaper</string>
       <string name="wallpaper_scrolling_label">Scroll Wallpaper</string>
       <string name="wallpaper_depth_effect_label">Wallpaper Depth Effect</string>
-      <string name="wallpaper_depth_effect_description">Zoom into and out of the wallpaper when transitioning between states.</string>
+      <string name="wallpaper_depth_effect_description">Zoom into and out of the wallpaper when transitioning between areas of the launcher.</string>
```
yasanglass, line 166: The previous string was more correct; this setting only toggles the wallpaper effect that happens when the notification panel is opened afaik and not the horizontal scroll animations/effects. "States" is not the best string but is more correct.

yasanglass (changes requested):
Only that one small change should be enough.

Also, I suggest splitting the changes into their own separate PRs instead of a single multi-purpose PR whenever possible.

lawnchair/res/values/strings.xml
SuperDragonXD: > this setting only toggles the wallpaper effect that happens when the notification panel is opened afaik

Well, the effect also happens when you go to the recents, open the app drawer, or swipe to the Google Feed. Basically changed the wording to that since some users are confused on what "states" meant.

I can't think of a different wording, so maybe you could suggest something else that's clear to other users.

yasanglass: Hmmm, my problem with the new string is that it says "areas of the launcher" but the effect only happens when you move outside the launcher. Maybe we can just mention recents & notification panel? to avoid confusing the users.

SuperDragonXD: I guess I could do that but it'll make the description longer

Also here is the video for that "zoom effect" (recents not included since I don't have QS):
 https://user-images.githubusercontent.com/70206496/187017185-749d4adc-b566-4400-b180-87ff83a21d12.mp4

suphon-t: > but the effect only happens when you move outside the launcher

The effect actually happens in the launcher: feed, app drawer, dragging icons (and recents if you have qs). So this wording makes sense. The notification panel also has this effect but that's not controlled by this toggle or the launcher.

yasanglass: I guess it's okay then, although I still wish we could find a way to make the string clearer. Since it happens in so many places, I don't think that is possible without way too many lines of text.

PR #2900, opened 2022-09-01
Change TopBar.kt to use LargeTopAppBar

SuperDragonXD:
## Description

This PR changes TopBar.kt to use LargeTopAppBar. Alongside that, it bumps `androidx.compose.material3` to `beta01` (needed for some functionality to work, alongside some minor changes to support that version), and removes `rememberFloatingState()` due to it being unused.

## Type of change

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -358,7 +358,7 @@ dependencies {
     implementation "androidx.compose.foundation:foundation:$compose_version"
     implementation "androidx.compose.material:material-icons-extended:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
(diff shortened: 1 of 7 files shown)


yasanglass:
Haven't fully looked into the code (yet) but the current implementation doesn't have the proper behavior; You cannot go back to the initial state by scrolling to the top. Although it sometimes happens in a glitchy way as you can see in the video.

Peek 2022-09-04 15-47

SuperDragonXD:
Fixed that behavior (see most recent commit). Do you want me to make the title larger when the top bar is uncollapsed?

PR #2989, opened 2022-10-01
Fix top bar scroll behavior

suphon-t:
## Description

Previously, we were temporarily using the `enterAlwaysScrollBehavior` behavior for the top bar, which expands when scrolling up even when not at the top.
This PR fixes the bug in the scroll stretch effect that interferes with the top bar expansion logic, and changes the scroll behavior of the top bar to `exitUntilCollapsedScrollBehavior`.

Before:
[before.webm](https://user-images.githubusercontent.com/8080853/193392374-1c2f0d81-0006-44bb-bf7f-073308c0ba87.webm)

After:
[after.webm](https://user-images.githubusercontent.com/8080853/193392379-29c2f88e-440f-4001-b219-0bf7db16a3d3.webm)

## Type of change

✅  Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/components/NestedScrollStretch.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/components/NestedScrollStretch.kt
@@ -121,7 +121,7 @@ private class NestedScrollStretchConnection(context: Context, invalidate: Runnab
                 bottomEdgeEffect.onRelease()
             }
         }
-        return available
+        return Offset.Zero
     }
 
     override suspend fun onPreFling(available: Velocity): Velocity {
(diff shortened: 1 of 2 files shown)


SuperDragonXD (approved):
Thanks paphomb!

SuperDragonXD:
Just realized this now @paphonb:

On Pre-S devices, the default overflow effect shows alongside the custom stretch effect as seen in the below video. Passing a `bottomBar` removes the effect though.

https://user-images.githubusercontent.com/70206496/193448944-e4c75f58-ef2b-470e-9ff1-b7b1973137a4.mp4

Was thinking of just setting [`LocalOverscrollConfiguration` to null](https://stackoverflow.com/questions/69468212/remove-lazycolumn-overscroll-effect-in-jetpack-compose) to fully hide any scroll effects, but you may have a better solution though.

PR #2991, opened 2022-10-01
Color selector improvements

suphon-t:
## Description

Since we switched the color selector from a bottom sheet to a full screen page, it makes sense to combine options from the "Dynamic" and "Presets" pages and make use of the available vertical space.

Also did some clean up on the color preference itself. We now have a source of truth for color preference definitions.

## Type of change

✅  General change (non-breaking change that doesn't fit the below categories like copyediting)

### DIFF
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -46,7 +46,7 @@
     <string name="app_info_drop_target_label">App Info</string>
     <string name="debug_restart_launcher">Restart Lawnchair</string>
     <string name="experimental_features_label">Experimental Features</string>
-    
+
     <!-- ExperimentalFeaturesPreferences -->
     <!-- <string name="experimental_features_label" />-->
     <string name="font_picker_label">Font Customization</string>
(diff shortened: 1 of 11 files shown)


kahanzammi:
In Presets tab, why don't you add a title to the basic color section. 
Example: Basic Colors

yasanglass:
About the UI changes, the original image looks a bit confusing. The mock-up by @kahanzammi looks much better but it still feels a bit too much.

Looks good other than the UI changes imo.

yasanglass (approved):

SuperDragonXD (approved):
Looks good, although I agree with Yasan on adding the headers as seen in the mockup above for better organization. (Maybe change the headers in there as "Dynamic" and "Swatches", though)

Goooler (approved):

suphon-t:
Added the section headers and a few new UX changes. Should look good now.

yasanglass:
Now both tabs have the same exact style! Looks good! 👌🏼

SefaMeert:
System launcher support on Android 13

SefaMeert:
Will there be system launcher support for Android 13?

PR #3033, opened 2022-10-15
Add permission ACCESS_RESTRICTED_SETTINGS

zjunior06:
This will allow modify restricted settings on Android 13

## Description

Fixes #(issue) 

## Type of change

✅ New feature (non-breaking change which adds functionality)

### DIFF
--- a/AndroidManifest-common.xml
+++ b/AndroidManifest-common.xml
@@ -43,6 +43,7 @@
     <uses-permission android:name="android.permission.READ_DEVICE_CONFIG" />
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.EXPAND_STATUS_BAR" />
+    <uses-permission android:name="android.permission.ACCESS_RESTRICTED_SETTINGS" />
 
     <!-- for rotating surface by arbitrary degree -->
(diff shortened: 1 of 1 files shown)


suphon-t:
Does this actually have any effect? I don’t see the permission being mentioned in the documentation.

zjunior06:
> Does this actually have any effect? I don’t see the permission being mentioned in the documentation.

Yes, it as, already tested and working.

zjunior06:
Before:
Screenshot_20221015-052431


After:
Screenshot_20221015-052918_Settings

yasanglass (approved):

daywalk3r666:
This don't work. Action build with this commit has no access to the permission.

zjunior06:
> This don't work. Action build with this commit has no access to the permission.

You need allow it, App Info > 3-dots

daywalk3r666:
Right but then you don't need this commit. Lawnchair alpha 4 can do this without this commit

zjunior06:
> Right but then you don't need this commit. Lawnchair alpha 4 can do this without this commit

But new builds can't. (Dev builds)

yasanglass:
@zjunior06 Thank you for contributing to _Lawnchair_!

PR #3049, opened 2022-10-23
At a Glance (smartspace) provider selection support

yasanglass:
## Description

Adds a system that allows users to select the provider of smartspace. The current options are:
1. Lawnchair: The built-in & customizable _At a Glance widget_
2. Google: The _At a Glance_ widget from _the Google app_
3. Google Search: The search bar widget from _the Google app_

The ultimate goal of this feature is not to let users choose between these 3 options as they are already accessible using normal widgets: The goal is to implement a system that can be used later on to introduce new smartspace providers easily. 

## More information
- The two google-powered ones are disabled when the Google app is not installed (or if it is installed but not enabled). 
- If the active smartspace provider is no longer _available_, the active smartspace automatically reverts back to _LawnchairSmartspace_ which is **always** available.
- The two google-powered ones technically keep working even without the Google app but they barely provide any functionality and are pretty much useless, that's why I chose to **not** let users use these smartspace providers without the Google app.
- After some discussion with the core Lawnchair team, I decided to show the name "At a Glance provider" to the user but use "Smartspace Mode" in the actual code as SmartspaceProvider class already exists in the code. Let me know if you think we can name them more clearly.

smartspace-provider

## Type of change

:x: General change (non-breaking change that doesn't fit the below categories like copyediting)
:x: Bug fix (non-breaking change which fixes an issue)
:white_check_mark: New feature (non-breaking change which adds functionality)
:x: Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #3065, opened 2022-10-30
Cleanup exception catching under lawnchair

Goooler:
## Description

Follow up https://github.com/LawnchairLauncher/lawnchair/pull/3049#discussion_r1002692600

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
@@ -189,8 +189,9 @@ class LawnchairLauncher : QuickstepLauncher(), LifecycleOwner,
 
         if (prefs.autoLaunchRoot.get()) {
             lifecycleScope.launch {
-                runCatching {
+                try {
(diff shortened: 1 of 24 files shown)


yasandev:
I'm not that familiar with `runCatching`, but isn't this bad practice? It looks like it just catches any exception and it's bad practice to just catch all exceptions. It's always better to catch specific exceptions, that's why multiple different exceptions are caught in some of the original code.

Other than the syntax improvement, this doesn't look like a positive change.

Only a few of the original catches catch all exceptions. Maybe we should just update those few.

yasanglass (changes requested):
I'm not that familiar with `runCatching`, but isn't this bad practice? It looks like it just catches any exception and it's bad practice to just catch all exceptions. It's always better to catch specific exceptions, that's why multiple different exceptions are caught in some of the original code.

Other than the syntax improvement, this doesn't look like a positive change.

Only a few of the original catches catch all exceptions. Maybe we should just update those few.

Goooler:
I think so, I'm reverting those changes now.

yasanglass (approved):

PR #3125, opened 2022-11-23
Group app list by sectionName before sorting for Simplified Chinese only

### LINKED ISSUE #2986: [Feature] Sort of Chinese and English
### Describe the feature

# Engligh
> About the sorting problem of Chinese and English
# Chinese
> 关于中英文的排序问题

# English
> Separate sorting makes me often need to swipe to the bottom to click apply(eg Google, Gmail, YouTube)
# Chinese
> 分离的排序使我常常需要滑动到底部来点击应用（例如Google，Gmail，YouTube）

No

### Example(s)

#  About the sorting problem of Chinese and English
### Additional context

# Wrong sort
- 电话
- Chrome
# Correct ordering
- Chrome
- 电话

Goooler:
## Description

Sort the app list to be more localized, before vs after:

Closes #2986

## Type of change

### DIFF
--- a/src/com/android/launcher3/allapps/AppInfoComparator.java
+++ b/src/com/android/launcher3/allapps/AppInfoComparator.java
@@ -24,6 +24,7 @@ import com.android.launcher3.pm.UserCache;
 import com.android.launcher3.util.LabelComparator;
 
 import java.util.Comparator;
+import java.util.Locale;
 
 /**
  * A comparator to arrange items based on user profiles.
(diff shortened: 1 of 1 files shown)


Goooler:
@yasandev @SuperDragonXD @paphonb 
Could you test this feature with your local languages?

yasanglass:
Doesn't improve sorting for Persian. All my apps with Persian names are at the bottom of the list.

Goooler:
So apps with Persian names should be sorted and mixed in English names, right? Could you show me the `sectionName` in `mApps` look like?

yasanglass:
image

Yeah, it should be mixed. Niagara Launcher does sort them properly. That ت symbol is the equivalent of T and should be sorted as T.

Goooler:
I see your `sectionName` list contains non-ascii chars, but not in my case, so the key point is the `computeSectionName`.

Goooler:
> Doesn't improve sorting for Persian. All my apps with Persian names are at the bottom of the list.

Is it in the same order as before?

yasanglass:
> Is it in the same order as before?

I think so.

Goooler:
I've tested this change in Japanese env, `sectionName`s contain non-ascii chars too, so can I preserve this change for Chinese only?

Goooler:
It's hard to adopt every language, can I preserve this change for Simplified Chinese only?

yasanglass:
> It's hard to adopt every language, can I preserve this change for Simplified Chinese only?

Yeah, I suppose its better than nothing. I can look into improving sorting for Persian and Arabic later myself as I need it for my own projects as well.

yasanglass (approved):

PR #3154, opened 2022-12-07
Fix to detect older lawnicons version as themed icon pack

naveenccmsd:
## Description

Currently older version of lawnicons not detected as the new themed intent is added only to new versions, this change is provide backward compatibility to older version of lawnchair untill 1.2.0.

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/IconPackPreferences.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/IconPackPreferences.kt
@@ -221,8 +221,7 @@ private fun getIconPackItemWidth(
     while (true) {
         gutterCount += 1f
         visibleItemCount += 1f
-        val possibleIconPackItemWidth =
-            (availableWidth - gutterCount * gutterWidth) / visibleItemCount
+        val possibleIconPackItemWidth = (availableWidth - gutterCount * gutterWidth) / visibleItemCount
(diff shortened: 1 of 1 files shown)


SuperDragonXD (approved):

ghost:
Lawnicons will be lost if the lawnchair restarts for some reason

naveenccmsd:
@arm64-v9a can you give version no for lawnchair and lawnicon

ghost:
> @arm64-v9a can you give version no for lawnchair and lawnicon

https://github.com/LawnchairLauncher/lawnchair/actions/runs/3648775645
https://github.com/LawnchairLauncher/lawnicons/actions/runs/3646451402

naveenccmsd:
Tried both version lawnaicons style has applied , Even after restart,  Did you mean lawnicons not shown in Icons pack list??

ghost:
> Tried both version lawnaicons style has applied , Even after restart, Did you mean lawnicons not shown in Icons pack list??

IMG_20221209_133924
After restarting lawnchair, themed icons will turn off again.

But I changed the signatures of lawnchair Dev and lawnicons Dev, the performance of Debug version is very low.  I generally use the Release version I build myself.  Since QuickSwitch turns lawnchair into a system app, switching the original signed Debug APK would be cumbersome.  Would a different signature affect this?

PR #3244, opened 2023-01-13
Add an option to override max search result count in settings

Goooler:
## Description

Would like to show more search results, but limited by the `LawnchairAppSearchAlgorithm.maxResultsCount`, we can override this value in preferences.

Follow up #3236.

## Type of change

### DIFF
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -322,6 +322,7 @@
       <string name="pref_search_auto_show_keyboard">Automatically Show Keyboard</string>
       <string name="fuzzy_search_title">Fuzzy Search</string>
       <string name="fuzzy_search_desc">Approximate matching for app searches.</string>
+      <string name="max_search_result_count_title">Max search result count</string>
 
     <!-- <string name="grid" /> -->
       <string name="app_drawer_columns">App Drawer Columns</string>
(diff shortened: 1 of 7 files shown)


lawnchair/res/values/strings.xml
```diff
@@ -322,6 +322,7 @@
       <string name="pref_search_auto_show_keyboard">Automatically Show Keyboard</string>
       <string name="fuzzy_search_title">Fuzzy Search</string>
       <string name="fuzzy_search_desc">Approximate matching for app searches.</string>
+      <string name="max_search_result_count_title">Max search result count</string>
```
SuperDragonXD:
```suggestion
      <string name="max_search_result_count_title">Max Search Result Count</string>
```

Use Title Case, as said in README.

SuperDragonXD (commented):
I'd recommend changing it to number of rows instead of number of items, since it's unlikely that users will choose 7 max search items for example.

(This can simply be done by multiplying the inputted number by app drawer column count)


Otherwise the PR's fine.

Goooler:
Will do later.

suphon-t (approved):

Goooler:
Need to input `InvariantDeviceProfile.GridOption` as a param if we use `PreferenceManager2.drawerColumns` in `LawnchairAppSearchAlgorithm`, it is not necessary, let's keep the `maxSearchResultCount` instead of `maxSearchResultRow`?

yasanglass (approved):

PR #3255, opened 2023-01-17
Distinguish between release and debug via `applicationIdSuffix` & `label`

Goooler:
## Description

To divide the debug with the release, easy to test the debug app instead of covering the release one installed.

## Type of change

### DIFF
--- a/quickstep/res/values/strings.xml
+++ b/quickstep/res/values/strings.xml
@@ -18,9 +18,6 @@
 -->
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
 
-    <!-- Application name -->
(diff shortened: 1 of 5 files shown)


Goooler:
One more thing, if we should use `Lawnchair debug` as the app name in debug mode?

quickstep/res/values/strings.xml
```diff
@@ -18,9 +18,6 @@
 -->
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
 
-    <!-- Application name -->
-    <string name="derived_app_name" translatable="false">Quickstep</string>
```
Goooler, line 22: Need to remove this due to the res conflicts.

suphon-t (approved):

PR #3275, opened 2023-01-21
Release apk with uploading mapping file

Goooler:
## Description

We can easily retrace obfuscated stack traces with the mapping file.

https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions

## Type of change

### DIFF
--- a/.github/workflows/build_release_apk.yml
+++ b/.github/workflows/build_release_apk.yml
@@ -38,7 +38,9 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: Release APK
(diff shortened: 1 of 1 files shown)


Goooler:
Not sure if we need to update here:

https://github.com/LawnchairLauncher/lawnchair/blob/95272645e3c05f45b7c2d93d3e6123a37a266074/.github/workflows/build_release_apk.yml#L61-L73

SuperDragonXD (approved):
Regarding your question, I don't think it's needed, unless if we need to upload the mapping file on the Telegram CI channel.

PR #3291, opened 2023-01-31
Rewrite overly large SQL queries

Goooler:
## Description

https://developer.android.com/reference/androidx/room/RewriteQueriesToDropUnusedColumns

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/data/iconoverride/IconOverrideDao.kt
+++ b/lawnchair/src/app/lawnchair/data/iconoverride/IconOverrideDao.kt
@@ -14,9 +14,11 @@ interface IconOverrideDao {
     @Query("DELETE FROM iconoverride WHERE target = :target")
     suspend fun delete(target: ComponentKey)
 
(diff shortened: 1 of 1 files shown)


suphon-t:
The entity contains [2 columns](https://github.com/LawnchairLauncher/lawnchair/blob/8dcd4eb32583c9e9298d01f9c758925bd93b5c25/lawnchair/src/app/lawnchair/data/iconoverride/IconOverride.kt), all of which used by the query. The query is nowhere near "overly large".

Goooler:
Yep, noticed that. This might be a change to prevent those "overly large".

yasanglass:
Are any of the columns actually unused? And there are only 2 columns as @suphon-t mentioned. Probably won't cause any real optimization & I am not sure how much it'll affect the compile time as it is done on compile.

Goooler:
There are the same results in their generated DAO implementations, I can close this one until we really need these optimizations.

PR #3723, opened 2023-11-14
Add Spotless to check style

### LINKED ISSUE #2844: Intro ktlint to check Kotlin style?
### About this discussion

Use plugins that support Ktlint (like [Kotlinter](https://github.com/jeremymailen/kotlinter-gradle)) to check and format Kotlin code.

Goooler:
## Description

Closes #2844.

## Type of change


lawnchair/src/com/google/android/libraries/launcherclient/LauncherClient.java
```diff
@@ -20,9 +20,11 @@
 import android.view.Window;
 import android.view.WindowManager;
 import android.view.WindowManager.LayoutParams;
+
+import java.lang.ref.WeakReference;
+
```
yasanglass: What rule is making this change? I don't understand how this is better than the previous style.

Goooler: This is auto-formatted by Studio when I was fixing wildcard imports. We can revert these changes in Java or set up Java format for them by Spotless.

yasanglass: We cannot have java formatting checks unless its only applied to the Lawnchair module. Otherwise we will get tons of conflicts due to the formatting changes compared to AOSP. If possible, you should make spotless only look at the Lawnchair module as well. 

About this change, I am fine with it as long as it is actually the default formatting configuration on Android Studio.

Goooler: I meant just apply styles for files under lawnchair folder.

yasanglass: That would be nice but I don't think we have much Java code in the Lawnchair folder so I wouldn't spend time on it.

yasanglass (approved):

suphon-t (approved):

PR #3773, opened 2023-11-26
Tweak CI configs

Goooler:
## Description

- Verify release build for every commit or PR, so that we can find potential obfuscation issues.
- Merge `build_debug_apk.yml` with `build_release_apk.yml`.
- Rename `build_debug_apk.yml` to `ci.yml`.

## Type of change


yasanglass:
This will disable the debug warning on CI builds. While technically the builds will no longer be debug, they are still unofficial releases & we should show a proper warning to the users. Maybe we can introduce a new build type that is optimized but will show the warning?

Goooler:
I can revert the change of merge `release-build`, but seems `build_release_apk.yaml` is unnecessary, we just need to keep `ci.yml` and https://github.com/LawnchairLauncher/lawnchair/blob/13-dev/.github/workflows/release_update.yml

PR #3791, opened 2023-11-29
Dynamic enable/disable GestureNavContract

### LINKED ISSUE #3762: [FEATURE] Use SEPARATE-RECENT-ACTIVITIE which will add closing animation for apps
### Describe the feature

In launcher3 i found code which has separate recent activity which adds the closing animations for custom launchers adding this feature would improve launcher experience and using closing animations without quick switch

The launcher experience would be enhanced for non rooted users

No

### Example(s)

Screenshot_20231124-193845
This can be found in developer options if possible

### Additional context

Maybe this feature was used in launchers such as hyperion and nova this should work with android 12 devices but not with i think android 11 lower

### LINKED ISSUE #3784: [FEATURE] `GestureNavContract` support
### Describe the feature

Technically a dupe of #2856, although it was closed as stale a while back and I thought I might open an issue focusing a more specific and realistic implementation for rootless users.

Since Android 11 or so, many launcher3-based system launchers and some OEMs such as Samsung have offered the [`GestureNavContract`](https://cs.android.com/android/platform/superproject/+/master:packages/apps/Launcher3/src/com/android/launcher3/GestureNavContract.java) class to third party launchers as a method of providing better animations to users without the need for something like QuickSwitch.

While not being exactly the same, the animations are an improvement over the current "scale-to-the-middle" effect that apps do by default. Lawnchair currently doesn't seem to use this API, although other launchers such as nova, hyperion, and niagra have implemented animations using this API to [pretty good success](https://twitter.com/prjkt_io/status/1551634957030010881).

Offers a better experience/opening and closing animations for rootless users and those who do not use quickswitch.

No

### Example(s)

https://github.com/LawnchairLauncher/lawnchair/assets/42101043/978662bf-79f7-4a97-a545-fb1386818b06

https://twitter.com/prjkt_io/status/1551634957030010881

### Additional context

Before A13, there was a [bug](https://issuetracker.google.com/issues/202864574) where the animations tended to end early, although this has been fixed in AOSP and by many OEMs following Android 13's release.

MrSluffy:
## Description

Introducing this enable/disable dynamically `GestureNavContract` feature provides  non-rooted users with the opportunity to enjoy a sophisticated gesture-based app-to-home navigation experience through the `GestureNavContract` without breaking for Quickswitch users.

Demo 

https://github.com/LawnchairLauncher/lawnchair/assets/36076410/5dba1623-b6bb-43ff-a4ef-e91e813c8b58

Closes #3784 and #3762

## Type of change

### DIFF
--- a/lawnchair/res/values/config.xml
+++ b/lawnchair/res/values/config.xml
@@ -104,6 +104,7 @@
     <bool name="config_default_show_icon_labels_in_drawer">true</bool>
     <bool name="config_default_enable_fuzzy_search">false</bool>
     <bool name="config_default_enable_smartspace">true</bool>
+    <bool name="config_default_separate_recent_activity">false</bool>
     <bool name="config_default_enable_feed">true</bool>
     <bool name="config_default_enable_icon_selection">false</bool>
     <bool name="config_default_show_component_names">false</bool>
--- a/lawnchair/res/values/config.xml
+++ b/lawnchair/res/values/config.xml
@@ -104,7 +104,6 @@
     <bool name="config_default_show_icon_labels_in_drawer">true</bool>
     <bool name="config_default_enable_fuzzy_search">false</bool>
     <bool name="config_default_enable_smartspace">true</bool>
-    <bool name="config_default_separate_recent_activity">false</bool>
     <bool name="config_default_enable_feed">true</bool>
     <bool name="config_default_enable_icon_selection">false</bool>
     <bool name="config_default_show_component_names">false</bool>
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -60,6 +60,8 @@
     <string name="transparent_background_icons">Transparent Themed Icons</string>
     <string name="transparent_background_icons_description">Use transparent background on themed icons</string>
     <string name="smartspace_mode_selection">At a Glance Provider selection</string>
+    <string name="pref_separate_recents_activity_title">Separate Recents Activity</string>
+    <string name="pref_separate_recents_activity_description">Uses a separate recents activity instead of using the integrated recents+Launcher UI (GesturNavContract API)</string>
 
     <!-- GeneralPreferences -->
     <string name="home_screen_rotation_label">Home Screen Rotation</string>
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -60,8 +60,8 @@
     <string name="transparent_background_icons">Transparent Themed Icons</string>
     <string name="transparent_background_icons_description">Use transparent background on themed icons</string>
     <string name="smartspace_mode_selection">At a Glance Provider selection</string>
-    <string name="pref_separate_recents_activity_title">Separate Recents Activity</string>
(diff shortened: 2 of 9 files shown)


lawnchair/res/values/strings.xml
```diff
@@ -60,6 +60,8 @@
     <string name="transparent_background_icons">Transparent Themed Icons</string>
     <string name="transparent_background_icons_description">Use transparent background on themed icons</string>
     <string name="smartspace_mode_selection">At a Glance Provider selection</string>
+    <string name="pref_separate_recents_activity_title">Separate Recents Activity</string>
```
SuperDragonXD: Switch preference options should ideally be in `<verb>` `<what to do>`. Also no one will need to know the internal code term, so I've clarified it as well by using "GestureNavContract":

```suggestion
    <string name="pref_separate_recents_activity_title">Enable GestureNavContract</string>
```

lawnchair/res/values/strings.xml
```diff
@@ -60,6 +60,8 @@
...
+    <string name="pref_separate_recents_activity_description">Uses a separate recents activity instead of using the integrated recents+Launcher UI (GesturNavContract API)</string>
```
SuperDragonXD: Users will have no idea what GestureNavContract means, so I've also clarified the description.
```suggestion
    <string name="pref_separate_recents_activity_description">Improves app animations by interfacing with the system launcher. Not all OEMs support it.</string>
```

SuperDragonXD (changes requested):
Looks fine, although I'd suggest clarifying the wording.

I'd also recommend hiding the option if QuickSwitch is available (via adding a `LawnchairApp.isRecentsAvailable` check).

suphon-t:
Can't we just change the condition [here](https://github.com/LawnchairLauncher/lawnchair/blob/f696cf85556ef1f875afef9bd8b250ed80ac8442/quickstep/src/com/android/launcher3/uioverrides/QuickstepLauncher.java#L628) instead of adding a toggle? I feel like that would be much cleaner, and would work without the user changing any settings.

MrSluffy:
> Can't we just change the condition [here](https://github.com/LawnchairLauncher/lawnchair/blob/f696cf85556ef1f875afef9bd8b250ed80ac8442/quickstep/src/com/android/launcher3/uioverrides/QuickstepLauncher.java#L628) instead of adding a toggle? I feel like that would be much cleaner, and would work without the user changing any settings.

Let me try

lawnchair/res/values/config.xml
```diff
@@ -104,6 +104,7 @@
     <bool name="config_default_show_icon_labels_in_drawer">true</bool>
     <bool name="config_default_enable_fuzzy_search">false</bool>
     <bool name="config_default_enable_smartspace">true</bool>
+    <bool name="config_default_separate_recent_activity">false</bool>
```
SuperDragonXD: Do we still need this? I think this can be safely removed.

SuperDragonXD (approved):

lawnchair/res/values/config.xml
MrSluffy: yeah let me remove it

suphon-t (approved):

PR #3799, opened 2023-11-30
Fix DT2S (Root) using keyevent

### LINKED ISSUE #2910: [BUG] Double tap to sleep no longer works through root access
### Describe the bug

Hello,

as written in the title, I recently installed a freshly compiled update for lawnchair, and now the double tap to suspend in homescreen doens't work anymore through root. I have to enable the accessibility service, and suspending this way is slower and a little bit glitchy (screen turns on for a second after locking).

Thanks for your support

### Steps to reproduce

1. Enable "suspend" function in gestures
2. Double tap to suspend

### Expected behavior

I expect the screen to shut off instantly, as it was using root.

Now instead it required to enable accessibility service; then, after I double tap, the screen locks, turns on for a second, and turns off again.

### Screenshots

_No response_

### Device information

Device: Poco F2
OS: Android 12L, CrDroid 8.8
Lawnchair Version: 12.1.0 Dev #1063

### Additional context

_No response_

MrSluffy:
## Description

Temporary fixed for root DT2S using keyevent

Fixes #2910

## Type of change


lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
```diff
@@ -63,12 +63,11 @@ class SleepGestureHandler(context: Context) : GestureHandler(context) {
 }
 
 class SleepMethodRoot(context: Context) : SleepGestureHandler.SleepMethod(context) {
-    private val rootHelperManager = RootHelperManager.INSTANCE.get(context)
 
-    override suspend fun isSupported() = rootHelperManager.isAvailable()
+    override suspend fun isSupported() = Shell.getShell().isRoot
```
Goooler, line 67: Why don't use `rootHelperManager.isAvailable()` here?

MrSluffy: it's not working anymore for some reason, I'm not sure tho what's causing it

MrSluffy: Also already tried to fix the current dt2s but this what I get, instead of make it sleep it just redirect to lockscreen

**Using old implementation :** 

https://github.com/LawnchairLauncher/lawnchair/assets/36076410/7f16a463-32aa-4bb9-84dd-80bf8f1a6787

Goooler: https://github.com/LawnchairLauncher/lawnchair/blob/7aa0b7172c75000bfd3e113cc009f52dab9f979a/lawnchair/src/app/lawnchair/root/RootHelperManager.kt#L27

Try
```kt
    private val isAvailableDeferred by lazy {
        scope.async(Dispatchers.IO) {
            Shell.getShell().isRoot
        }
    }
```

MrSluffy: I think it's better to drop it, why access it if we can simplify it like the latest commit one?

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
```diff
@@ -63,12 +63,11 @@ class SleepGestureHandler(context: Context) : GestureHandler(context) {
...
 
     override suspend fun sleep(launcher: LawnchairLauncher) {
-        rootHelperManager.getService().goToSleep()
+        Shell.su("input keyevent 26").exec()
```
Goooler, line 70: And we can put this into `RootHelperManager` and wrap it as a suspend fun like:
```kt
    suspend fun exec(command: String) = withContext(Dispatchers.IO) {
        Shell.getShell().newJob().add(command).exec()
    }
```

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
Goooler: Not sure why, I guess it might be a blocking job, better to put it on IO?

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
MrSluffy: Yes it's fancy, but why are we doing this?, is there any use cases for this that we can use for? or just for future one? if this for future, I think let's keep it minimal what you think? and apply it if there's other uses cases

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
MrSluffy: No, it will work by changing IO to Main

but if we gonna use `Shell.su("input keyevent 26").exec()` for DTS2 no need to put in RootHelperManager or you want to keep the current one, you will get output like the screen record I send above

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
Goooler: https://github.com/topjohnwu/libsu?tab=readme-ov-file#shell-operations

`exec` is synchronous, `submit` is async.

MrSluffy: Yes we only need to execute keyevent 26 once and no job other than allowing root permision, since it's use Shell.su it will ask root permission if magisk, if kernelSu just manually allow root permision in kernelsu app

Goooler (approved):

lawnchair/src/app/lawnchair/gestures/handlers/SleepGestureHandler.kt
```diff
@@ -63,12 +63,11 @@ class SleepGestureHandler(context: Context) : GestureHandler(context) {
 }
 
 class SleepMethodRoot(context: Context) : SleepGestureHandler.SleepMethod(context) {
-    private val rootHelperManager = RootHelperManager.INSTANCE.get(context)
 
-    override suspend fun isSupported() = rootHelperManager.isAvailable()
+    override suspend fun isSupported() = Shell.getShell().isRoot
 
     override suspend fun sleep(launcher: LawnchairLauncher) {
```
Goooler, line 69:
```suggestion
    override suspend fun sleep(launcher: LawnchairLauncher) = withContext(Dispatchers.IO) {
```

Goooler:
One more thing, we should check accessibility permission and use it first, fallback to root access if there is no accessibility granted.

PR #3862, opened 2023-12-16
Add live information feature

yasanglass:
## Description

Adds support for dynamically updating some content on the app. 

- [x] Load the feature config from resources

## Type of change


MrSluffy (commented):
This looks great. Please make it dismissable either swiping or x button.

Also this is great we can link this in GitHub action too in debug builds so user will notify too when new debug builds is available

lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/LiveInformationRequest.kt
```diff
@@ -0,0 +1,49 @@
...
+        if (response.isSuccessful) {
+            val responseBody = response.body()?.string() ?: return@withContext null
+
+            val liveInformationObject = JSONObject(responseBody)
+
+            val liveInformation = LiveInformation(
+                announcements = Announcement.fromJsonArray(
+                    liveInformationObject.getJSONArray("announcements"),
+                ),
+            )
```
suphon-t: Why are we parsing the JSON manually? Can we use a Retrofit converter instead?

lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/model/Announcement.kt
```diff
@@ -0,0 +1,53 @@
...
+                            url = it.optString("url", ""),
+                            active = it.optBoolean("active", true),
+                            test = it.optBoolean("test", false),
+                        )
+                    }
+                    .toImmutableList()
+            } catch (e: Exception) {
+                Log.e("Announcement", "fromJsonArray: Failed to parse: ${e.message}")
+                persistentListOf()
+            }
```
suphon-t: Same with this. We already have `kotlinx.serialization`, so why not use it

suphon-t:
Also agree with Sluffy about dismissing announcements. We will probably need announcement ids tho.

lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/LiveInformationRequest.kt
yasanglass: Because I thought we don't use them. This code is inspired by the search result code by @MrSluffy. I did check if we use Moshi anywhere in the app though.

I would use Moshi but it is not used in Lawnchair anywhere else so far. As you mentioned on your other comment we should probably use `kotlinx.serialization` instead of this. I will look into updating both requests.

MrSluffy: We made an exception due to differences of the data response, in startpage the response is like this `["popcorn", [ "popcorn", "popcorn instrumental" ] ]` as we don't want to serialize and map it afterwards since it doesn't take a key value pair, so we use a straightforward approach.

lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/model/Announcement.kt
yasanglass: Discussing on https://github.com/LawnchairLauncher/lawnchair/pull/3862#discussion_r1429191716

lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/LiveInformationRequest.kt
yasanglass: Fixed now!

yasanglass:
> Also agree with Sluffy about dismissing announcements. We will probably need announcement ids tho.

IMO it's not that important but it all depends on how we will use the announcements. I will work on it. The UI aspect of it is a bit vague for me now but I probably have to do it via horizontal swipes.

MrSluffy (approved):

PR #3912, opened 2023-12-23
Tweak ProGuard rules

Goooler:
## Type of change

### DIFF
--- a/proguard.pro
+++ b/proguard.pro
@@ -42,13 +42,14 @@
 -keep class app.lawnchair.LawnchairApp { *; }
 -keep class app.lawnchair.LawnchairLauncher { *; }
 -keep class app.lawnchair.compatlib.** { *; }
+
 -keep class com.google.protobuf.Timestamp { *; }
+
 # TODO: Remove this after the change in https://github.com/ChickenHook/RestrictionBypass/pull/9 has been released.
 -keep class org.chickenhook.restrictionbypass.** { *; }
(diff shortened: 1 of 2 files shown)


proguard.pro
```diff
@@ -42,13 +42,14 @@
 -keep class app.lawnchair.LawnchairApp { *; }
 -keep class app.lawnchair.LawnchairLauncher { *; }
 -keep class app.lawnchair.compatlib.** { *; }
+
 -keep class com.google.protobuf.Timestamp { *; }
+
 # TODO: Remove this after the change in https://github.com/ChickenHook/RestrictionBypass/pull/9 has been released.
 -keep class org.chickenhook.restrictionbypass.** { *; }
--keep class androidx.compose.runtime.** { *; }
```
Goooler, line 48: Is there any specific reason to keep all classes under this path?

MrSluffy: Yes, or you can verify building in release?

Goooler: I don't see any crash in Settings screen after removing this line, where you saw?

MrSluffy: No, I mean you successfully build in release build right? And no crash?. I forgot why I added, but if you don't faced any issue in release build that's great. 

Anyway will check it

Goooler: I checked, nothing crashed, just keep the removing now.

MrSluffy (approved):

PR #4044, opened 2024-02-03
Sort app results first if they start with the query

### LINKED ISSUE #4038: [BUG] Undesirable search sorting in app drawer
### Describe the bug

The current searching is nice as we don't have to enter the full correct app name, but when searching for an app, the results should at least sort apps starting with the input name first, and not alphabetically

Its really annoying if i just enter the app name and have to enter much more than would be necessary otherwise. This is happening whether fuzzy is on or off, but much more noticable if its on.

### Steps to reproduce

1. Open app drawer search
2. start typing an app name
3. see some apps that are not starting with the entered 

For example i have to enter _insta_ in order to have Instagram on first position, although i don't have any other apps starting with _ins_ an while typing i and n Instagram doesn't show up at all.

### Expected behavior

Apps that start with the entered query should be first.

### Screenshots

i
in
ins
inst
insta

### Device information

- Device: Samsung Galaxy A33
- OS: stock OneUI 6.0, Android 13
- App version: 12.1.0 Alpha 4

### Additional context

_No response_

neuos:
## Description

Search results are now sorted additionally if the app name starts with the query.
Can be seen as feature or bug fix depending on perspective, but should be just a small change with no larger impact.

Fixes #4038

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/search/LawnchairAppSearchAlgorithm.kt
+++ b/lawnchair/src/app/lawnchair/search/LawnchairAppSearchAlgorithm.kt
@@ -173,6 +173,7 @@ class LawnchairAppSearchAlgorithm(context: Context) : LawnchairSearchAlgorithm(c
         return apps.asSequence()
             .filter { StringMatcherUtility.matches(queryTextLower, it.title.toString(), matcher) }
             .filterHiddenApps(queryTextLower)
+            .sortedWith(queryComparator(queryTextLower))
             .take(maxResultsCount)
             .toList()
     }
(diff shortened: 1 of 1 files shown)


MrSluffy (approved):
Thank you for your contribution

Goooler:
This breaks fuzzy search results.

MrSluffy:
> This breaks fuzzy search results.
> 
> 

What's the expect result?

Goooler:
As we can see, none of the apps contains "ww" in their `sectionName + name`.

Goooler:
Reverted this, if you still want to fix #4038 may need another way.

PR #4599, opened 2024-07-26
Implementation home screen layout

### LINKED ISSUE #4323: [FEATURE] Drawer-less mode
### Describe the feature

A drawer-less mode which not only hides the drawer completely, but also forces all icons to be displayed and arranged on the homescreen (unless manually hidden).

Many users that came from iOS 13 and below, MIUI/HyperOS, etc. may like a drawer-less launcher instead of drawers to gain more control of app arrangement. However, no well-known third party launcher today provides this feature.

A drawer-less mode may not take much work to be added, which should mainly reuse the auto-add-to-homescreen code, and add a little logic to (re-)sync with the all apps and hidden apps list.

Switching between the drawer-less mode and the original mode can be simplified to just prompting for confirmation, (recommending a backup?), and resetting the homescreen arrangement totally.

No

### Example(s)

[Lawndesk](https://github.com/j67678/Lawndesk) was a Lawnchair O/P fork to achieve this, whose codebase is now outdated and no longer maintained.

### Additional context

_No response_

MrSluffy:
## Description

Introducing the home screen layout feature! Now, users have the ability to personalize their experience by disabling the app drawer and arranging apps directly on their home screen. With the innovative flowerpot implementation, users can effortlessly categorize their apps, adding a touch of organization and aesthetic appeal to their device interface.

- Closes : #4323

## TODO
- Own DB

## Type of change


rikumi:
I tried out the test build from your branch and here's a quick report of how this feature works now:

- The 'experimental feature' entry is fine and works as I expected. However this should be a breaking change as most of the homescreen functionalities would work in different ways.
- The flowerpot folders keep disappearing and being recreated from time to time.
- Icons keep resetting to the left-most position and does not respect user arrangements.
- Icons that are dragged out of the flowerpot folders get duplicated within the folder on the next launch.

Technically, I think the flowerpot folders and new icons should get saved into the database once they're generated, in order to be treated equally as user placed icons. And only the previously unseen icons should be automatically added. The current implementation might need further considerations.

Kallinteris-Andreas:
It puts each app in its own "desktop pane" (not sure what it is actually called), instead of in a folder named "other" or putting them all in the same panes 


https://github.com/user-attachments/assets/21e5a316-e02a-44eb-9057-b60daa7e8c59

MrSluffy:
> It puts each app in its own "desktop pane" (not sure what it is actually called), instead of in a folder named "other" or putting them all in the same panes
> 
>  screen-20240912-112417+.mp4

it's still WIP and as if you can see the PR is still draft!

PR #4659, opened 2024-08-16
Fix edge-to-edge on three-button nav bar

Goooler:
## Description

| Before | After |
|--------|--------|
| before | after |

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/PreferenceActivity.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/PreferenceActivity.kt
@@ -18,42 +18,23 @@ package app.lawnchair.ui.preferences
 
 import android.content.Context
 import android.content.Intent
-import android.graphics.Color
 import android.os.Bundle
-import androidx.activity.SystemBarStyle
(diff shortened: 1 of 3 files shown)


lawnchair/src/app/lawnchair/ui/preferences/PreferenceActivity.kt
```diff
@@ -18,42 +18,23 @@ package app.lawnchair.ui.preferences
...
 class PreferenceActivity : AppCompatActivity() {
     @OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        enableEdgeToEdge()
         setContent {
-            val windowSizeClass = calculateWindowSizeClass(this)
-            val darkTheme = isSelectedThemeDark
-
-            DisposableEffect(darkTheme) {
```
Goooler, line 43: No need to use `DisposableEffect` here, we don't handle `onDispose`.

SuperDragonXD (approved):

PR #4775, opened 2024-09-06
Mark stable classes and enable Compose Compiler reports

Goooler:
## Description

https://developer.android.com/develop/ui/compose/performance/stability/fix#configuration-file

## Type of change


yasanglass:
@Goooler does this provide any benefit other than being able to drop a dependency and not having to use immutable list class? I have not used this method before and am wondering about possible side-effects.

Goooler:
> Compose treats immutable collections as stable, now we can mark normal collections stable instead. This will also ensure Compose's smart skipping works efficiently and reduce our package size.

_Originally posted by @Goooler in https://github.com/LawnchairLauncher/lawnicons/issues/2280#issuecomment-2333769920_

yasanglass (approved):

PR #4792, opened 2024-09-11
Firefox and iceraven addition in qsb

Unofficial-Life:
- Added Firefox and Iceraven to QSB
- Monet icons also added for the above
- please do correct any typos if found


Unofficial-Life:
due to not knowing how github works i might have by mistake added more commits will make an updated pull request

Issue #4663, opened 2024-08-17
[FEATURE] Larger range for App Drawer 'Row height' (lower minimum value)


tgex0:
### Describe the feature

I'd like a lower minimum value for App Drawer 'Row height'.
I have a grid of 8 columns of icons (at 70% size) in the app drawer and find that the smallest 'Row height' of 70% still leaves a lot of space between the rows.
Having experimented with different values, I've found that a 'Row Height' of 50% works great for this scenario.

### Why would it be useful to add?

Gives the user greater flexibility in configuring the app drawer.

### Did the feature exist in Lawnchair v2? (Play Store version)

No

### Example(s)

Current lowest value of 70%:

![Screenshot_2024-08-17-22-51-30-250_app lawnchair](https://github.com/user-attachments/assets/97d4fb4c-2688-4df9-b64f-c606e3b71632)

With lower value of 50%:

![Screenshot_2024-08-17-22-51-40-930_app lawnchair](https://github.com/user-attachments/assets/3dc27640-7df5-4394-88ea-bf181df970b1)

charliefrance:
How did you change it to 50%? 70% is much too large, I want to match it with my 7x15 homescreen

tgex0:
> How did you change it to 50%? 70% is much too large, I want to match it with my 7x15 homescreen

I forked the project, changed the source and built a new APK.
All I did was change the slider's range in the UI (_lawnchair\src\app\lawnchair\ui\preferences\destinations\AppDrawerPreferences.kt_).
I have no idea if it'll have any unwanted knock-on effects but it seems to work fine for me.

ghost:
I've just upgraded to 0.14 beta 3 and it seems to have changed since beta2 - I'd never touched the row height before, but the App Drawer felt instantly different in how much space was between icons vertically. I have 2x Pixel 6a (stock ROMs), the one running the stock Pixel launcher still has a tighter row height than Lawnchair 0.14beta3 at 70%.

To my untrained eye, a setting of 60% or 50% allowed in Lawnchair row height should/would match the default Android 14 Pixel 6a row height density when holding my two devices side by side, using a column count of 5 wide. $0.02

[...]

PR #4795, opened 2024-09-11
Fix floating point arithmetic error

tgex0:
## Description

Fixes root cause of this crash: https://github.com/LawnchairLauncher/lawnchair/issues/4663#issuecomment-2308932350
Floating point arithmetic rounding causes error in comparison when checking a UI slider's range/steps.

Crash was avoided by changing lower range of slider to 0.3f but using a value of 0.2f causes the app to crash.
Changing a different slider's range value could potentially cause a similar crash in the future so fixing the root cause of the problem is necessary.

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/components/controls/SliderPreference.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/components/controls/SliderPreference.kt
@@ -140,9 +140,9 @@ fun getSteps(valueRange: ClosedFloatingPointRange<Float>, step: Float): Int {
     if (step == 0f) return 0
     val start = valueRange.start.toBigDecimal()
     val end = valueRange.endInclusive.toBigDecimal()
-    val test = (end - start) / step.toBigDecimal()
-    val steps = test.toInt()
-    require(test.compareTo(steps.toBigDecimal()) == 0) {
+    val decimalSteps = (end - start) / step.toBigDecimal()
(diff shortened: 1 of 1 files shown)


lawnchair/src/app/lawnchair/ui/preferences/components/controls/SliderPreference.kt
```diff
@@ -138,10 +138,11 @@ fun SliderPreference(
...
     if (step == 0f) return 0
-    val start = valueRange.start
-    val end = valueRange.endInclusive
-    val steps = ((end - start) / step).toInt()
-    require(start + step * steps == end) {
+    val start = valueRange.start.toBigDecimal()
+    val end = valueRange.endInclusive.toBigDecimal()
+    val test = (end - start) / step.toBigDecimal()
+    val steps = test.toInt()
+    require(test.compareTo(steps.toBigDecimal()) == 0) {
```
SuperDragonXD: I think using a more descriptive variable name aside from `test` could be helpful. Here's a suggestion:


```suggestion
    val decimalSteps = (end - start) / step.toBigDecimal()
    val steps = decimalSteps.toInt()
    require(decimalSteps.compareTo(steps.toBigDecimal()) == 0) {
```

tgex0:
Done!

SuperDragonXD (approved):

Goooler (approved):

PR #4796, opened 2024-09-11
Fix calculator search precision and rounding errors

### LINKED ISSUE #4756: [BUG] Rounding/precision errors in calculator search
### Describe the bug

There are a few rounding/precision errors when using the search input as a calculator.

### Steps to reproduce

1. Try pi*0, result is 0E-15 (0 would be a more useful answer)
2. Try pi*1.000, result is 3.141592653589793000 (there are not three zeros in the decimal expansion of pi at that position)
3. Try 2.0*6.0, result is 6.00 (most calculators would just display 6)
4. Try1/3*3, result is 0.9999999999999999 (technically, an infinite number of 9s here _would_ equal 1 but 1 would be the expected result)

### Expected behavior

I would expect the results of the calculator search to be the same as that of any other calculator.

I expect the reason for (2) above is that pi is truncated/rounded to 15 decimal places before the calculation is performed. (This would apply to e as well.)
I expect the reason for (3) above (and also the reason for the trailing zeros in (2)) is that the precision for the result is increased, knowing that more decimal places might be required for the result.

### Screenshots

_No response_

### Device information

- Device: POCO X5 5G
- OS: HyperOS 1.0.5.0 EEA (latest stock)
- App version: Lawnchair.Debug.14-dev.Nightly-CI_1319-8f89197.apk

### Additional context

_No response_

tgex0:
## Description

Fixes calculator search result precision and rounding errors.

Previously:
- pi*0 = 0E-15
- pi*1.000 = 3.141592653589793000
- 2.0*6.0 = 6.00
- 1/3*3 = 0.9999999999999999

This PR fixes such issues, now resulting in:
- pi*0 = 0
- pi*1.000 = 3.141592653589793
- 2.0*6.0 = 6
- 1/3*3 = 1

Fixes #4756 

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/search/algorithms/data/Calculation.kt
+++ b/lawnchair/src/app/lawnchair/search/algorithms/data/Calculation.kt
@@ -1,6 +1,8 @@
 package app.lawnchair.search.algorithms.data
 
 import app.lawnchair.search.algorithms.data.calculator.Expressions
+import java.math.BigDecimal
+import java.math.MathContext
 
 data class Calculation(
     val equation: String,
@@ -12,9 +14,17 @@ fun calculateEquationFromString(
     query: String,
 ): Calculation {
     return try {
-        val result = Expressions()
-            .eval(query)
-            .toString()
+        val evaluatedValue = Expressions().eval(query)
+        val roundedValue = evaluatedValue.round(MathContext.DECIMAL64)
+        val formattedValue = roundedValue.stripTrailingZeros()
+        val absoluteValue = formattedValue.abs()
+        val threshold = BigDecimal("9999999999999999")
+
+        val result = if (absoluteValue.compareTo(threshold) > 0) {
+            formattedValue.toString()
+        } else {
+            formattedValue.toPlainString()
+        }
 
         Calculation(
             equation = query,
--- a/lawnchair/src/app/lawnchair/search/algorithms/data/Calculation.kt
+++ b/lawnchair/src/app/lawnchair/search/algorithms/data/Calculation.kt
@@ -20,7 +20,7 @@ fun calculateEquationFromString(
         val absoluteValue = formattedValue.abs()
         val threshold = BigDecimal("9999999999999999")
 
-        val result = if (absoluteValue.compareTo(threshold) > 0) {
+        val result = if (absoluteValue > threshold) {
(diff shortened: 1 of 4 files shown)


tgex0:
I changed this PR to a draft to fix a further problem that was revealed by my initial fix.

https://github.com/LawnchairLauncher/lawnchair/pull/4796/commits/a527e91948f9a9cc93908e8f691cb27eace008c4 was added to ensure that the equation displayed in the search result tile always reflects the actual search typed in by the user.

The content of the tile (the result and the equation) is only updated if the result or the id changes. (The id uses the result so if the result does not change, the id will not change either.)
Previously, if a user typed in 3\*1.0 and then added another zero, the result _would_ change (to 3.100) so the tile was updated and the problem was not evident.

After my initial commit https://github.com/LawnchairLauncher/lawnchair/pull/4796/commits/41ab80a5a2fa8ed8dc18e834308a1c293dcabff2, the result no longer shows insignificant trailing zeros so the result does not change (from 3\*1.0 to 3\*1.00, for example) and, thus, the tile is not updated if the user adds insignificant zeros to the search.

If the user were to type in 3\*1.00001, the equation in the tile would not be updated with all the zeros until the final 1 was entered.

Without changing the logic of when the tile should be updated, the easiest solution is to change the id (with a random value on each search input change), as the result itself cannot be changed.

SuperDragonXD (approved):

lawnchair/src/app/lawnchair/search/algorithms/data/Calculation.kt
```diff
@@ -12,9 +14,17 @@ fun calculateEquationFromString(
...
-        val result = Expressions()
-            .eval(query)
-            .toString()
+        val evaluatedValue = Expressions().eval(query)
+        val roundedValue = evaluatedValue.round(MathContext.DECIMAL64)
+        val formattedValue = roundedValue.stripTrailingZeros()
+        val absoluteValue = formattedValue.abs()
+        val threshold = BigDecimal("9999999999999999")
+
+        val result = if (absoluteValue.compareTo(threshold) > 0) {
```
Goooler:
```suggestion
        val result = if (absoluteValue > threshold) {
```

Goooler (approved):

PR #4798, opened 2024-09-11
Copy calculator search result to clipboard

### LINKED ISSUE #4729: [FEATURE] Calculator copy on interaction
### Describe the feature

It would be very convenient if tapping on the App Drawer's Search Calculator result would copy the result to clipboard.

Convenience, usability, speed.

No

### Example(s)

Don't have any code, but here's a [video](https://github.com/user-attachments/assets/71e13153-35c9-4a09-ac4a-caa0631bea87).

### Additional context

_No response_

tgex0:
## Description

New feature to copy the calculator search result to the clipboard (by just tapping the calculator search result tile).

Fixes #4729

## TODO
~1. Revert icon to regular _calculator.xml_ (reference in _SearchTargetFactory.kt_)~
~2. Add _ic_copy.xml_ icon to right-hand side of calculator search result tile~

## Type of change


tgex0:
Basic functionality works fine but there is no indication that tapping the tile will copy the result to the clipboard. Maybe this is fine as it is.

Rather than adding text somewhere or a further icon, I thought the least intrusive solution might be to change the calculator icon to a combo calculator/copy-to-clipboard icon. Something like this:

icon

Please let me know what you think. Thanks.

SuperDragonXD:
I'd suggest to add an icon to copy the result to the clipboard (at the right side), as it makes it significantly clear that you can copy the result.

tgex0:
I agree but, unfortunately, I could not figure out how to do that.

SuperDragonXD:
I guess we can make the icon a temporary solution as of now (make sure to mark it as `TODO`).

tgex0:
OK, I've added and implemented the new combo icon and added a TODO in this PR's description.

tgex0:
I'll leave this in draft mode as I need to fix the icon...

tgex0:
Icon is probably OK. I think the problem I experienced with it may be due to #4807.

tgex0:
Icon not displaying correctly (in Dark theme) is due to #4813. (No problem with icon. Problem was with backup file.)

tgex0:
TODO tasks complete.

SuperDragonXD (approved):
Hello again, and thanks for your patience.

While this solution isn't exactly elegant, it works for the purpose of showing single that the text is copyable. I'd probably refactor this (alongside other related code) after some time.

tgex0:
Just for clarification, I was able to include the regular clipboard icon at the right-hand side of the calculator results tile as you suggested (via a new layout) and the current PR reflects this. I reverted the calculator icon to the original one.

Screenshot

I hope that's OK.

MrSluffy:
Hi can you update this to `15-dev` Thanks!

MrSluffy (approved):

PR #3146, opened 2022-12-02
Migrate BaseActivity to AppCompat

Goooler:
## Description

Migrate the `BaseActivity` to `AppCompat` to better compat the changes in AndroidX libs in the future.

## Type of change

:white_check_mark: General change (non-breaking change that doesn't fit the below categories like copyediting)
:x: Bug fix (non-breaking change which fixes an issue)
:x: New feature (non-breaking change which adds functionality)
:x: Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #4803, opened 2024-09-12
Make BaseActivity extends ComponentActivity

Goooler:
## Description

We don't have to maintain ComponentActivity like behaviors in LawnchairLauncher anymore.

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
@@ -19,34 +19,14 @@ package app.lawnchair
 import android.app.ActivityOptions
 import android.content.Context
 import android.content.Intent
-import android.content.IntentSender
 import android.graphics.drawable.Drawable
 import android.os.Bundle
-import android.os.Handler
-import android.os.Looper
 import android.view.Display
 import android.view.View
 import android.view.ViewTreeObserver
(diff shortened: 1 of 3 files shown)


SuperDragonXD (commented):
Have you ensured that there aren't any major side-effects if we remove these? Seems like this PR could remove functionality that isn't available in `ComponentActivity`.

Goooler:
I land this in my fork for a long time.

I've tried to replace it with `AppCompatActivity` in #3146, but it might not be a good choice. Now we can try `ComponentActivity` instead, which supports Lifecycle and Compose but AppCompat things.

MrSluffy (approved):
This is great, please merge it before weekend since merge A15 WIP it's not much change in launcher3 legacy

edit : I haven't test this in old android, it shouldn't be a problem in error like ResourceId not found thing?

Goooler:
I guess resource id things should be related to AppCompat* instead of Component*.

suphon-t:
> Seems like this PR could remove functionality that isn't available in `ComponentActivity`.

The additions I wrote are mostly for backporting `ComponentActivity`'s features, so this shouldn't be a problem. What I'm worried about is the additional features in `ComponentActivity` that the regular `Activity` doesn't have. But as @Goooler mentioned it's working in his fork, so that should be fine too.

suphon-t (approved):

PR #4828, opened 2024-09-16
Fix Iceraven and Mull QSB provider website links

tgex0:
## Description

Changed Iceraven and Mull QSB provider website links to start with "https://".
All other QSB provider website links (without exception) start with "https://".

## Type of change

## Additional info

I realize that this will not fix an issue that was raised in the Telegram channel recently whereby the Iceraven and Mull download links in the UI take the user to a Google Play Store "Item not found." page but this is a small problem that ought to be fixed regardless.

### DIFF
--- a/lawnchair/src/app/lawnchair/qsb/providers/Iceraven.kt
+++ b/lawnchair/src/app/lawnchair/qsb/providers/Iceraven.kt
@@ -13,7 +13,7 @@ data object Iceraven : QsbSearchProvider(
     packageName = "io.github.forkmaintainers.iceraven",
     action = "org.mozilla.fenix.OPEN_TAB",
     className = "org.mozilla.fenix.IntentReceiverActivity",
-    website = "github.com/fork-maintainers/iceraven-browser/releases/latest",
+    website = "https://github.com/fork-maintainers/iceraven-browser/releases/latest",
     type = QsbSearchProviderType.APP,
     supportVoiceIntent = true,
 ) {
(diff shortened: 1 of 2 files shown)


Goooler (approved):

Unofficial-Life:
It still doesn't work. i am the person who reported the same on tg 

I have droidify which has both the apps selecting that solves the problem but as i mention on tg its structured as it will only redirect to a vendor app (play store, aurora store, etc any application like that).

tgex0:
Unfortunately, I knew this would not resolve the issue you brought up (see Additional Info above) but the links _did_ need the _https://_ protocol added to them anyway (for consistency with all other QSB providers), hence this PR.

The problem you refer to seems to be due to the _launchOnAppMarket_ function in lawnchair/src/app/lawnchair/qsb/providers/QsbSearchProvider.kt

The function never gets past the first 'try' as the 'Item not found.' page does not seem to trigger an error (for the 'catch' to come into play) and, anyway, the 'website' variable does not form part of this function so it would never be used, even if the 'try' failed.

I've tried to detect whether the package is available on the Google Play Store (before showing the 'Item not found.' page) using things like the HTTP header responses, Intents and the PackageManager class but was unable to come up with a satisfactory solution.

Maybe best to just log a new bug report (at least for the moment). Maybe someone more qualified than myself will be able to resolve it.
If a get anything working in the meantime, I'll open a new PR.

PR #4889, opened 2024-10-14
set enableEdgeToEdge() directly in Activity´s onCreate()

Rafael2616:
reference: https://developer.android.com/develop/ui/views/layout/edge-to-edge?hl=en#enable-edge-to-edge-display

this fixes the black nav bar in some older android devices form factors running api 28 this issue only affects dark theme in my device, in light theme the nav bar is transparent as espected. this commits fixes this by ensuring edge to edge is appled - tested and worked

## Description

Fixes #(issue) 

## Type of change

### DIFF
--- a/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairLauncher.kt
@@ -25,8 +25,8 @@ import android.os.Bundle
 import android.view.Display
 import android.view.View
 import android.view.ViewTreeObserver
-import androidx.activity.enableEdgeToEdge
 import androidx.activity.SystemBarStyle
+import androidx.activity.enableEdgeToEdge
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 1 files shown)


Rafael2616:
20241014_000954-COLLAGE

Rafael2616:
home screen:
20241014_002218-COLLAGE

Rafael2616:
hi @MrSluffy, i found an issue with edge to edge implementation, this is a long time issue in lower api devices, so i was able to solve it by adding this in LawnchairLauncher.kt onCreate()
```
        if (!Utilities.ATLEAST_Q) {
            enableEdgeToEdge(
                navigationBarStyle = SystemBarStyle.auto(
                    Color.TRANSPARENT,
                    Color.TRANSPARENT,
                ),
            )
        }
``` 
but i dont think that this is an good solution because it override the actual  EdgeToEdge implememntation defined in Theme.kt, thats why i close this pull. so this issue only affects dark mode in api 28 or lower in some form factors. but i saw that LawnchairLauncher.kt uses View system to construct the screen and I'm not very familiar, do you think it can be easily fixed??

MrSluffy:
that's great, Create an PR! Thanks!

MrSluffy:
Yes this is great, I'll merge it

Rafael2616:
ok, I'll reopen this and make the changes as in the comment

Rafael2616:
done

MrSluffy (approved):

PR #4973, opened 2024-11-06
Improve documentation v4 (LC15)

validcube:
## Description

This PR marked the fourth major change to the documentation.

close #4878 

~~Right now this PR needs updating the screenshot to Lawnchair 15 which isn't possible, I'm currently making LC15 demo-able before I have this PR finished.~~

Here's a list of tasks that need to be done to be marked as ready:

### Lawnchair 15 Preview
- [x] #4975 
- [x] #5029
- [x] Match Icon Shape (61c3db874e07a6b99fd2512aaab520a15086f7f8)
- [x] Update the preview

### Developer
- [x] How to update locally bundled Google Fonts listing
- [x] ~~Inform regarding WallpaperManager~~
- [x] Update for https://github.com/LawnchairLauncher/lawnchair/pull/5034

### Miscellaneous
- [x] #4878 

## Type of change

:white_check_mark: General change (non-breaking change that doesn't fit the below categories like copyediting)
:x: Bug fix (non-breaking change which fixes an issue)
:x: New feature (non-breaking change which adds functionality)
:x: Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #5029, opened 2024-12-01
Allow Launcher3 to use grid/hotseat/icon size value

validcube:
## Description

This is the same one used in #4973 (or well, it's not pushed)

~~Right now the icon doesn't scale up well, the preview for changing grid size properly size this.~~

## Type of change

✅ Bug fix (non-breaking change which fixes an issue)


validcube:
Ready, icon now properly scaled up

MrSluffy (approved):

PR #5112, opened 2024-12-22
Allow re-arranging popup menu items

SuperDragonXD:
## Description

This feature allows users to re-order and toggle specific items in the pop-up menu.

Below is a screenshot of the feature:

## Type of change


SuperDragonXD:
As a note: would prefer rebasing this to the dev branch (manually) instead of squash-merging. It's up to you guys though.

MrSluffy (approved):

istudyatuni:
There is a bug: when the home screen is locked, popup closes almost immediately after opening (not sure if it belongs to this PR, but I guess yes)

Also it seems that the old setting has not migrated: I had "Show lock button" turned on, but it was disabled in new settings

Also, I think it should be possible to enable/disable items like "Edit home screen" even when home screen is locked

SuperDragonXD:
> There is a bug: when the home screen is locked, popup closes almost immediately after opening (not sure if it belongs to this PR, but I guess yes)

Will likely need to move the check (when the home screen is locked) to the preference area; thanks for catching this issue.

> Also it seems that the old setting has not migrated: I had "Show lock button" turned on, but it was disabled in new settings

Preferences migration hasn't been implemented (yet). Will look into it when I have the time.

> Also, I think it should be possible to enable/disable items like "Edit home screen" even when home screen is locked

It doesn't really make sense that you are able to edit the home screen or add widgets while the home screen is locked, so I've purposefully disabled that option.

istudyatuni:
> It doesn't really make sense that you are able to edit the home screen or add widgets while the home screen is locked

Yes, but I'm talking about editing it in the added settings - you can't enable/disable this when home screen is locked. Previous behavior was the same (not able to toggle), but I think allowing this would be better. Maybe it's better to create a separate issue?

SuperDragonXD:
> Yes, but I'm talking about editing it in the added settings - you can't enable/disable this when home screen is locked. Previous behavior was the same (not able to toggle), but I think allowing this would be better. Maybe it's better to create a separate issue?

Allowing this behavior doesn't make sense as you *should not* be able to add or edit widgets when the home screen is locked. Same goes for re-arranging icons: if the home screen is locked, you *can not* go to edit mode and re-arrange the icons there.

Because of this, the items "Widgets" & "Edit home screen"  automatically turn off (get hidden) and disabled when the home screen is locked.

istudyatuni:
Hmm, okay

PR #5167, opened 2025-01-15
Fix upstream bug with dots enabled

thunder-coding:
## Description

I did not find any issue reported upstream, but I can confirm that this bug exists on Moto App Launcher (the launcher on Motorola's devices). Further there is an issue on GrapheneOS issue tracker, which seems to imply this is a bug in AOSP itself. I had asked @agnostic-apollo about this is a bug, and he said that this is a bug in Android's libraries itself. I'm not a Android dev, so feel free to suggest any changes to the PR.

With this change, I was able to use Lawnchair without any crashes for around 24+ hours with notification dots on. So, I guess this workaround works

Works around GrapheneOS/os-issue-tracker#2601

## Type of change


MrSluffy (approved):

MrSluffy:
LGTM! Thanks for the PR!

PR #5537, opened 2025-06-11
Fix crash on Android 16 beta caused by missing canImeRenderGesturalNavButtons()

BourbonFather:
Lawnchair crashes immediately on Android 16 beta (e.g., Pixel 8 Pro) due to a missing method in the Android framework:

```
java
InputMethodService.canImeRenderGesturalNavButtons()
```

This method is not present on some builds of Android 16, resulting in a `NoSuchMethodError` at launch.

Fix
Wrapped the method in `isImeRenderingNavButtons()` with a reflection-based check to safely detect and call it only when available. This prevents the launcher from crashing while preserving functionality on supported devices.

Fixes #5535

Type of change
:white_check_mark: Bug fix (non-breaking change which fixes an issue)

PR #5538, opened 2025-06-11
Fix: Adapt to missing Flags in android 16 qpr1 beta 2

### LINKED ISSUE #5535: [BUG]  Crash on Launch with Android 16  QPR1 beta 2
### Describe the bug

Had to force close Lawnchair to allow for change back to pixel launcher. Nightly, 14.3 Beta, and the version on Playstore all fail in a similar fashion

### Steps to reproduce

- start launcher, and immediately crashes
- fresh boot with Lawnchair as the default launcher causes immediate crash and you must force close in order to pick pixel launcher as the default

### Expected behavior

The way it's always worked, as my default launcher, working with smartspacer integration.

### Screenshots

_No response_

### Device information

Device: Pixel 9 pro XL
Android 16 QPR1 Beta 2

### App version

14 beta 3

### Additional context

_No response_

adil192:
## Description

Lawnchair immediately crashes after updating to Android 16 QPR1 beta 2 on my Pixel 9 Pro with [this error](https://github.com/LawnchairLauncher/lawnchair/issues/5535#issuecomment-2963598290):
`java.lang.NoClassDefFoundError: Failed resolution of: Landroid/appwidget/flags/Flags;`

I fixed it by copying the flag into an existing Flags file. Note that I copied it into systemui's Flags file, which is probably wrong/an antipattern. If this is unacceptable, it might be safe to just hardcode the flag as `true`, especially since the original flag was annotated with `@AssumeTrueForR8`.

Fixes https://github.com/LawnchairLauncher/lawnchair/issues/5535 for me, but note that (edit: some of) the bug reports in there point to a different error (referenced in #5537) for some reason? But this works fine on my phone at least

## Type of change

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ android.hideApiWarning=false
 kotlin.code.style=official
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx16g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
 org.gradle.parallel=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ android.hideApiWarning=false
 kotlin.code.style=official
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.jvmargs=-Xmx16g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
 org.gradle.parallel=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.parallel=true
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -8,7 +8,7 @@ android.hideApiWarning=false
 kotlin.code.style=official
 
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.jvmargs=-Xmx16g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
```
validcube:
```suggestion
org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8 -XX:+UseParallelGC
```

validcube (approved):
LGTM from me, looks familiar to how AOSP does it so it should be fine without bugs.

But I dislike that the overall project memory usage has been raised to 16 GB (from 4 GB), I would have prefer to keep the requirement more strict or as-is instead of laxing it.

SuperDragonXD (approved):
As validcube said, this PR looks good. Thanks for contributing to Lawnchair, @adil192!

PR #5573, opened 2025-06-26
Fix: Ensure swipe-down gesture detection is active for default action

goat040:
The VerticalSwipeTouchController was previously not enabling detection for swipe-down gestures if the configured action was the default (e.g., Open Notifications). This was due to the `overrideSwipeDown` flag only being true if the gesture was *different* from the default.

This commit modifies `getSwipeDirection()` to enable detection for swipe-down (and swipe-up) gestures whenever their configured action is not 'NoOp', regardless of whether it's the default or a custom setting. This allows the default "swipe down to open notifications" gesture to be correctly detected and handled.

Removed the now-redundant `overrideSwipeUp` and `overrideSwipeDown` fields and their associated update logic in the `init` block.

### DIFF
--- a/lawnchair/src/app/lawnchair/gestures/VerticalSwipeTouchController.kt
+++ b/lawnchair/src/app/lawnchair/gestures/VerticalSwipeTouchController.kt
@@ -2,18 +2,18 @@ package app.lawnchair.gestures
 
 import android.graphics.PointF
 import android.view.MotionEvent
-import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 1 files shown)


lawnchair/src/app/lawnchair/gestures/VerticalSwipeTouchController.kt
MrSluffy, line 1: just remove the commented code and it looks fine to me

MrSluffy (changes requested):

PR #5733, opened 2025-08-07
feat(ui): enhance overflow menu with custom animations and icons

foXaCe:
Replace standard DropdownMenu with custom Popup implementation featuring:
- Smooth scale and fade animations with custom timing curves
- Material 3 Card design with proper elevation and colors
- Leading icons for all menu items in preferences dashboard
- Improved visual hierarchy and consistency

before
before

after
after


SuperDragonXD (approved):
Hey, my apologies for the long delay in reviewing this. I've had a chance to go through it, and this is a solid improvement.

I've just pushed a few changes to your branch to align it with our current project standards before merging:

* I've changed the icons in the dropdown menu to be outlined (and rounded), to maintain visual consistency across the app.
* I've temporarily removed the custom `Popup` animations. You're right that they add a nice touch, but I ran into some edge-case bugs when trying to test them. Alongside that, we also plan to migrate to the newer `material3` library (which has Material 3 Expressive components), which comes with the new motion scheme animation system. Rather than maintain a custom implementation that will be replaced, it's better to stick with the default behavior for now and adopt the new M3 animations everywhere at once. I've kept the custom shape, as that's a great touch.

Merging this now. Thanks again for the contribution!

PR #6395, opened 2026-02-15
feat(workspace): Add "Set as default page" option

Exaper:
Description
                                                                                                                                                                                                                                
Adds a "Set as default page" option to the workspace long-press popup menu, allowing users to designate any workspace page as the default home page. Pressing HOME navigates to the chosen page instead of always returning to page 0. If the default page is deleted, it resets to page 0.

Reasoning

The default home page is hardcoded to page 0 (DEFAULT_PAGE = 0). Users who organize their workspace with a preferred page in the middle have no way to make HOME return there. This adds that ability through the existing long-press popup menu, requiring no new settings UI.

Testing

  1. Create 3+ workspace pages with content
  2. Long-press empty area on page 2 → verify "Set as default page" appears
  3. Tap it → verify toast "Default home page updated"
  4. Long-press on page 2 again → verify "Set as default page" does NOT appear (it's now the default)
  5. Long-press on page 0 or 1 → verify "Set as default page" DOES appear
  6. Navigate to page 0, press HOME → verify it scrolls to page 2
  7. Already on page 2, press HOME → verify gesture handler fires (same as current page-0 behavior)
  8. Set page 3 as default, remove all items from page 3 so it gets stripped, press HOME → verify fallback to page 0
  9. Clear app data → verify default behavior (page 0) is preserved

<img width="372" height="750" alt="image" src="https://github.com/user-attachments/assets/f9a36c0e-8f12-4abf-baaf-d51c1fdc900e" />

<img width="369" height="753" alt="image" src="https://github.com/user-attachments/assets/1ae08ba1-14c5-442a-9fb7-edca250995d4" />


  Type of change

  :x: Bug fix (A non-breaking change that fixes an issue)
  :white_check_mark: New feature (A non-breaking change that adds functionality)
  :x: Breaking change (A fix or feature that would cause existing functionality to not work as expected)
  :x: Refactor (A code change that neither fixes a bug nor adds a feature)
  :x: Performance (A code change that improves performance)
  :x: Style (Code style changes)
  :x: Docs (Changes to documentation)
  :x: Chore (Changes to the build process or other tooling)


src/com/android/launcher3/views/OptionsPopupView.java
```diff
@@ -213,6 +213,19 @@ public static ArrayList<OptionItem> getOptions(Launcher launcher) {
             OptionsPopupView::onWidgetsClicked,
             OptionsPopupView::startSettings
         );
+
+        // Show "Set as default page" only on non-default pages
+        if (!launcher.getWorkspace().isCurrentPageDefault()) {
```
SuperDragonXD: I'd suggest moving this logic to the `LauncherOptionsPopup` class, so we don't add unnecessary logic in this class. Since `launcher` is passed, we can also move the if logic to there.

src/com/android/launcher3/views/OptionsPopupView.java
```diff
@@ -279,6 +292,14 @@ private static boolean toggleHomeScreenLock(View v) {
...
     }
 
+    private static boolean setAsDefaultHomePage(View v) {
+        Launcher launcher = Launcher.getLauncher(v.getContext());
+        int currentPage = launcher.getWorkspace().getNextPage();
+        launcher.getWorkspace().setDefaultPage(currentPage);
+        Toast.makeText(launcher, R.string.default_home_page_set, Toast.LENGTH_SHORT).show();
+        return true;
+    }
+
```
SuperDragonXD: See above.

src/com/android/launcher3/views/OptionsPopupView.java
```diff
@@ -213,6 +213,19 @@ public static ArrayList<OptionItem> getOptions(Launcher launcher) {
             OptionsPopupView::onWidgetsClicked,
             OptionsPopupView::startSettings
         );
+
+        // Show "Set as default page" only on non-default pages
+        if (!launcher.getWorkspace().isCurrentPageDefault()) {
+            options.add(new OptionItem(
+                launcher,
+                R.string.set_default_home_page,
+                R.drawable.ic_home_screen,
```
SuperDragonXD: Instead of duplicating the previous icon, we should use something more suitable such as [`home_pin`](https://fonts.google.com/icons?selected=Material+Symbols+Rounded:home_pin:FILL@0;wght@400;GRAD@0;opsz@24&icon.style=Rounded&icon.query=home&icon.size=24&icon.color=%23e3e3e3&icon.platform=android).

lawnchair/src/app/lawnchair/preferences2/PreferenceManager2.kt
```diff
@@ -345,6 +345,11 @@ class PreferenceManager2 @Inject constructor(
         defaultValue = context.resources.getBoolean(R.bool.config_default_lock_home_screen),
     )
 
+    val defaultHomePage = preference(
+        key = intPreferencesKey(name = "default_home_page"),
+        defaultValue = 0,
```
SuperDragonXD:
```suggestion
        defaultValue = Workspace.DEFAULT_PAGE,
```

SuperDragonXD (changes requested):
Hello, and thanks for contributing to Lawnchair! Just have a few suggestions to improve the logic here.

src/com/android/launcher3/views/OptionsPopupView.java
Exaper: I wasn't sure if it's OK to bring in extra icons. Followed your suggestion and added a dedicated home pin icon (reflected in the screenshot).

SuperDragonXD (changes requested):
Thanks! One last thing: I'd suggest allowing the user to re-order or disable this option by adding it to `optionsList` and having a new item in `DEFAULT_ORDER`, i.e. `LauncherOptionPopupItem("default_page", true)`. You can then add the conditional logic afterwards as a `.filter {}` block near the end of the function.

We should also tweak `restoreMissingPopupOptions()` to add the missing items to the end of the list instead as well.

Exaper:
@SuperDragonXD Done! Added "default_page" to `DEFAULT_ORDER`, `optionsList`, and `getMetadataForOption()` so the option is fully "re-orderable" and can be toggled by the user like other popup items. The visibility condition (hide when already on the default page) is now a `.filter` in the chain. Also updated `restoreMissingPopupOptions()` to append missing items at the end (`currentOptions + missingItems`) instead of prepending them.

SuperDragonXD (approved):

SuperDragonXD:
Thanks, @Exaper! Merging this now.

PR #6419, opened 2026-02-23
fix(ui): resolve unnecessary launcherPopupOrder write

strangr:
### Description

When Launcher is created, reload grid is force called:
Everytime `LawnchairLauncher`  `onCreate` is triggered, `LauncherOptionsPopup.restoreMissingPopupOptions` is being called. inside it is unconditionally writing into `launcherPopupOrder`, that is triggering `reloadHelper.reloadGrid`.

This creates flicker, that first loads grids and suddenly reload it again.

### Reasoning

we don't need to writing into preferences if nothing has changed

### Testing

This should happen every time launcher is created

### Type of change


SuperDragonXD (approved):
Thanks, @strangr! Merging this now.

PR #6420, opened 2026-02-23
fix(model): resolve unnecessary model reload

### LINKED ISSUE #6274: [BUG]  pressing home button
### Describe the bug

When you are on any page other than the main home screen and push the home button the screen flashes but then stays on the same page. It should go back to page one. Using on a Samsung Galaxy s24

### Steps to reproduce

Go to page 2 and tap the home screen button, or gesture

### Expected behavior

Return to the main home screen

### Screenshots

_No response_

### Device information

Samsung Galaxy s24

### App version

15 beta 2

### Additional context

_No response_

strangr:
### Description

Fix unnecessary model reload every time model validation happens

Fixes #6274

### Reasoning

Every time model reload happens, we are checking if `mAppsList.hasShortcutHostPermission()` and `hasShortcutsPermission(mContext)` are equal (inside `ModelDelegate.java`'s validateData).

problem is that on some devices (Happens on my Galaxy S phone), some users apps are inaccessible, thus 
forcing model reload.

problem is that is never set correctly because we never reach that part of the code

### Testing

Happens on devices that have locked user with inaccessible apps list. For example galaxy phones with Secure Folder

### Type of change

### DIFF
--- a/src/com/android/launcher3/model/LoaderTask.java
+++ b/src/com/android/launcher3/model/LoaderTask.java
@@ -16,6 +16,7 @@
 
 package com.android.launcher3.model;
 
+import static android.os.Process.myUserHandle;
 import static com.android.launcher3.BuildConfigs.WIDGET_ON_FIRST_SCREEN;
 import static com.android.launcher3.Flags.enableLauncherBrMetricsFixed;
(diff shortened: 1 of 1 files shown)


SuperDragonXD (approved):
Thanks, @strangr!

PR #6469, opened 2026-03-03
fix app drawer search algorithm setting freeze

Mattwmaster58:
I wasn't able to change the search provider for some strange reason (
Settings -> Search -> App Drawer  -> Search Algorithm), it froze up everything.

**Full disclosure**, I had opencode/opus 4.6 churn on this for probably a cumulative 3-4h with a real device before it came up with the text below which *did* fix the issue but is very hard for me to determine whether this is the correct fix. I don't think this fix is correct but it might be useful for reference. Here's what it wrote:

--- 
Context

When LauncherPreviewRenderer's output view is embedded inside a Compose AndroidView (e.g. in TwoTabPreferenceLayout's HorizontalPager), onViewCreated() was calling view.rootView.setViewTreeLifecycleOwner(), which overwrote the entire window's lifecycle owner. When the pager disposed the preview page, onViewDestroyed() moved the lifecycle to DESTROYED, killing the host Compose tree and causing a freeze/blank screen.

Fix by setting the lifecycle owner on the view itself rather than rootView, so the BaseContext lifecycle is scoped to just the preview view and doesn't interfere with the host window's lifecycle.

### Testing

Settings -> Search -> App Drawer  -> Search Algorithm

### Type of change

### DIFF
--- a/src/com/android/launcher3/util/BaseContext.kt
+++ b/src/com/android/launcher3/util/BaseContext.kt
@@ -86,8 +86,8 @@ constructor(base: Context, themeResId: Int, private val destroyOnDetach: Boolean
             object : OnAttachStateChangeListener {
 
                 override fun onViewAttachedToWindow(view: View) {
-                    view.rootView.setViewTreeLifecycleOwner(this@BaseContext)
-                    view.rootView.setViewTreeSavedStateRegistryOwner(this@BaseContext)
+                    view.setViewTreeLifecycleOwner(this@BaseContext)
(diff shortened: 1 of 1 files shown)


strangr:
Hi, I had same issue and arrived to same fix.
just fyi, I didn't touch it because I was not sure that other classes that inherit BaseContext should also have same change. And since I had no way of testing those cases I left it alone.

this is just a heads up

Mattwmaster58:
@strangr that's exactly why i'm leaving this as a draft for now. I hope to at some point take a closer look to see if there's a better way to do this.

PR #6566, opened 2026-03-29
feat: Text classifier search provider

validcube:
### Description

Proof of Concept search provider using native Android ML Text Classifier models.

Available for Android 9 and above.

All done on-device, and no apk size increases as this is already in almost all of Android device

### Reasoning

Smarter search suggestions,
* Type a phone number, offer you to call, or message, add them to contact
* Type a email, offer to open the mail app
* Type a URL, offer to open browser
* Type a address, offer to open map
* Type a flight number, offer to track it
* Type a date, offer to show calendar on that date
* Type a time, offer to show calendar on that time, or create an event on that time
* Type a word from other languages, offer to translate it
* Type a currency, offer to convert it

The suggestions will show the app intent app icon, and open it.

### Testing

Testing environment used during development:
* 📱 Google Pixel 7 | Android 17.2 Beta 1 (Android 17 QPR2 Beta 1)
* 📱 Nothing (3a)-series | Android 16.0
* 📱 Nothing (3a)-series | Android 15

Search:
* 111-111-1111: Phone + SMS + Add contact
* 01/01/2026: Calendar
* 4pm: Calendar + Add calendar event
* 11:00: Calendar + Add calendar event
* https://example.com: Browser
* example@example.com: Email
* Generic address: Maps
* Generic flight number: Tracking
* Any languages that isn't system locale: Translate
* 149 THB: Currency convert

### Type of change

## Summary by CodeRabbit

* **New Features**
  * Adds text-classifier “microresults” to search on supported Android versions.
  * Adds settings to enable results and configure email, phone, address, URL, date/time, flight, and other categories.
  * Clarifies labels for action suggestions and web suggestions.
* **Bug Fixes**
  * Improves search-result activation by prioritizing PendingIntent actions.
  * Refines grouped backgrounds for adjacent search results.
* **Chores**
  * Adds a suggestions icon.


lawnchair/res/values/strings.xml
```diff
@@ -896,6 +896,7 @@
     <string name="search_pref_result_web_title">Web suggestions</string>
     <string name="search_pref_result_history_title">Search history</string>
     <string name="all_apps_search_result_calculator">Calculator</string>
+    <string name="search_pref_result_textclassifier">Smart suggestions</string>
```
validcube: Alternative naming are below:

```suggestion
    <string name="search_pref_result_textclassifier">Text suggestions</string>
```

```suggestion
    <string name="search_pref_result_textclassifier">Entity extractions</string>
```

MrSluffy (approved):

SuperDragonXD (commented):
I think it would be better if we somehow integrate this with other search providers for better customizability. As currently, this does conflict with other providers, providing similar results for the same query.

If that's not possible, another term for this would be "Microresults", similar to Google's integration.

validcube:
> I think it would be better if we somehow integrate this with other search providers for better customizability. As currently, this does conflict with other providers, providing similar results for the same query.

I've checked, it seems like it isn't possible to integrate it with other search providers or at least I haven't found a use case with the current APIs yet.

> If that's not possible, another term for this would be "Microresults", similar to Google's integration.

If I think correctly, you're saying that we rename the feature name to microresult instead right?

PR #6584, opened 2026-04-03
fix: cap oversized preview bitmaps with extreme aspect ratios

### LINKED ISSUE #6579: fix: Thumbnail sampler leaves oversized wide/tall images uncapped in search preview
## Summary

The `calculateInSampleSize` function in `lawnchair/src/app/lawnchair/search/adapter/SearchTargetFactory.kt` uses an `AND`-based loop condition that only increases the sample size while **both** dimensions exceed the requested bounds. This means the smaller edge is preserved at >= 256 px, but the long edge can remain far above the cap.

**Examples of problematic inputs:**
- `3000×1000` still decodes around `1500×500`
- `12000×120` does not downsample at all

This undermines the memory/CPU reduction intended by the image preview optimisation introduced in #6570.

## Proposed fix

Change the loop condition to use `OR` so that both axes are independently constrained:

```kotlin
while (rawHeight / inSampleSize > reqHeight || rawWidth / inSampleSize > reqWidth) {
    inSampleSize *= 2
}
```

This ensures the decoded bitmap always fits within `MAX_PREVIEW_SIZE_PX` on **both** dimensions.

## References

- PR: https://github.com/LawnchairLauncher/lawnchair/pull/6570
- Original review comment: https://github.com/LawnchairLauncher/lawnchair/pull/6570#discussion_r3010788713
- Requested by: @validcube

wellorbetter:
### Description

  `calculateInSampleSize` in `SearchTargetFactory.kt` used `&&` in its while-loop condition, meaning images with extreme aspect
  ratios (e.g. 12000×120) were never downsampled because only one dimension exceeded the cap. Switching to `||` ensures both axes are
   independently constrained, so the decoded bitmap always fits within `MAX_PREVIEW_SIZE_PX` on every dimension.

  Fixes #6579

  ### Reasoning

  The `&&` condition exits the loop as soon as either dimension drops below the threshold, leaving the other potentially far above
  the cap. Using `||` keeps looping until both dimensions are within bounds.

  ### Testing

  1. Trigger a search result containing an image with an extreme aspect ratio (e.g. very wide banner or tall thin image)
  2. Verify the preview is correctly downsampled without excessive memory usage

  ### Type of change

## Summary by CodeRabbit

* **Bug Fixes**
  * Thumbnails with invalid, non-positive, excessively large, or extreme aspect-ratio dimensions are now skipped and show the default icon to prevent decode failures.
  * Image downscaling logic refined to select sampling earlier and more reliably, reducing failed thumbnail loads and peak memory use.

### DIFF
--- a/lawnchair/src/app/lawnchair/search/adapter/SearchTargetFactory.kt
+++ b/lawnchair/src/app/lawnchair/search/adapter/SearchTargetFactory.kt
@@ -403,6 +403,15 @@ class SearchTargetFactory(
 
 object FilesTarget {
     private const val MAX_PREVIEW_SIZE_PX = 256
+    private const val MAX_RAW_DIMENSION_PX = 16_384
+    private const val MAX_ASPECT_RATIO = 5
+
+    private fun isValidPreviewSize(width: Int, height: Int): Boolean {
+        if (width <= 0 || height <= 0) return false
+        if (maxOf(width, height) > MAX_RAW_DIMENSION_PX) return false
+        if (maxOf(width, height) > minOf(width, height) * MAX_ASPECT_RATIO) return false
+        return true
+    }
 
     fun getPreviewIcon(
         context: Context,
@@ -421,6 +430,10 @@ object FilesTarget {
         return try {
             val options = BitmapFactory.Options().apply { inJustDecodeBounds = true }
             BitmapFactory.decodeFile(path, options)
+
+            val w = options.outWidth
+            val h = options.outHeight
+            if (!isValidPreviewSize(w, h)) return null
 
             options.inSampleSize = calculateInSampleSize(
                 options.outWidth,


wellorbetter (commented):
The `&&` condition exits the loop as soon as either dimension drops below the threshold, leaving the other potentially far above
  the cap. Using `||` keeps looping until both dimensions are within bounds.

lawnchair/src/app/lawnchair/search/adapter/SearchTargetFactory.kt
```diff
@@ -499,7 +499,7 @@ object FilesTarget {
             val halfHeight = rawHeight / 2
             val halfWidth = rawWidth / 2
             // Then we loop until we find the right sample size by the power of 2
-            while (halfHeight / inSampleSize >= reqHeight && halfWidth / inSampleSize >= reqWidth) {
+            while (halfHeight / inSampleSize >= reqHeight || halfWidth / inSampleSize >= reqWidth) {
```
validcube: In my opinion, I don't it is worth it to preview images in extreme aspect ratio scenario, the resulting preview would be vague to see or make out visually.

I'd suggest making a function that return a placeholder image if the initial image size has aspect ratio more than 5:1 and 1:5.

I'd also suggest returning a placeholder if image's height or width is more than 16,384px (theoretically maximum output of S23 Ultra raw) so that unusually large image won't be rendered by Lawnchair.

validcube (changes requested):

lawnchair/src/app/lawnchair/search/adapter/SearchTargetFactory.kt
wellorbetter: Added isValidPreviewSize guard per your suggestion — extreme aspect ratios (>5:1) and oversized images (>16384px) now fall back to the placeholder icon.

validcube (approved):

PR #6629, opened 2026-04-12
fix: Extend LawnchairApp with LauncherApplication

### LINKED ISSUE #6347: [BUG] App drawer doesn't update when installing or removing apps until launcher restarts
### Describe the bug

The app drawer does not immediately update when new apps are installed or existing apps are removed. The changes only become visible after manually restarting the launcher.

### Steps to reproduce

1. Install a new app from the Play Store or an APK.

2. Open the app drawer and notice the new app is missing.

3. Alternatively, uninstall an app and notice its icon remains in the app drawer.

4. Restart the launcher (via settings or force stop).

5. See that the app drawer now correctly reflects the changes.

### Expected behavior

The app drawer should automatically refresh and reflect newly installed or uninstalled apps immediately without requiring a manual restart of the launcher.

### Screenshots

_No response_

### Device information

Device: Redmi 12C
OS: LineageOS 23.0
App version: Gooler's release of Lawnchair 16

### App version

Lawnchair 16 (d70543025aceca434116226c4cb03bc0ddb0aa0d)

### Additional context

_No response_

### LINKED ISSUE #6366: [BUG] Uninstalled apps don't get removed from home screen and app drawer
### Describe the bug

When uninstalling an app, the icon won't disappear from the home screen and the app drawer. You can manually remove the app from the home screen, but you can't remove it from the app drawer. Tapping the uninstalled app results in a "App not installed." popup.

### Steps to reproduce

1. Tap and hold on an app (in my case Florid from F-Droid)
2. Select "Uninstall" and uninstall the app
3. Check the home screen and app drawer for the uninstalled app

### Expected behavior

Uninstalled apps get removed from the home screen and app drawer, installed apps appear in the app drawer and home screen, if the respective setting to add new apps to the home screen is enabled.

### Screenshots

https://github.com/user-attachments/assets/0df1d594-e987-4adc-884f-a43e4878c997

### Device information

- Device: Google Pixel 9a
- OS: GrapheneOS 2026020601 (Android 16)

### App version

https://github.com/LawnchairLauncher/lawnchair/commit/3b6ab31 (16.Dev.(#3406) Nightly)

### Additional context

_No response_

### LINKED ISSUE #6608: [BUG] App drawer do not update with new apps after the apps installs - not updated after app removal
### Describe the bug

The new install apps do not appear in app drawer as expected.      

The only solution is force stop lawnchair, so the app drawer get updated on restart.  

Steps to reproduce:  
1. open fdroid or play google and install app and note the name of app.     
2. check the app drawer and you will not find the new installed app.          

I have to force stop lawnchair and restart the lawnchair to see the new app. 

I also see another issue.   
The app that I removed will not disappear in app drawer
unless I do force stop to get the app drawer updated.      
https://github.com/LawnchairLauncher/lawnchair/issues/6366

### Steps to reproduce

for installs of new apps: 
 
1. open fdroid or play google and install app and note the name of app.     
2. check the app drawer, and you will not find the new installed app.         

removals of apps: 
1. uninstall/remove one of the apps.  
2.  check apps drawer, and you will still see the name of removed app.   

### Expected behavior

I should see new app in the app drawer right after I installed.  

I should see removed app disappear in the app drawer.  

### Screenshots

This is right after I installed Pokebase app.   Where is Pokebase app?   
Pokebase can be found on fdroid app.    

Image

I had to force stop the lawnchair to see Pokebase app.  
Image

### Device information

Device: Pixel 10 Pro XL  
OS: latest android 16    

### App version

16.Dev  latest commit - I built the apk 

### Additional context

_No response_

Mattwmaster58:
Fixes #6347, fixes #6366, fixes #6608 
### Description

Edit: approach completely changed thanks to the prodding of @strangr 

LawnchairApp now extends LauncherApplication, rather than Application. LauncherApplication is pretty thin anyway, and critically we now hit the path were we correctly initialize dagger:

https://github.com/LawnchairLauncher/lawnchair/blob/1d49f086122ba4acddd7b27507cf6e75e6795f51/src/com/android/launcher3/dagger/LauncherComponentProvider.kt#L37

This brings us closer in line with how Launcher3 works as well, so maybe some evidence of a good direction. I've also removed the code in LauncherProvider.onCreate, since LauncherApplication handles that for us now.

I don't have a good answer for why this was never done before and know the FULL implications of such a change. It seems like it was somewhat intentional, this comment would have been irrelevant (we would never hit the paths the engage safe mode) IF this approach was taken from the start, so clearly someone was aware of the approach being taken:
https://github.com/LawnchairLauncher/lawnchair/blob/1d49f086122ba4acddd7b27507cf6e75e6795f51/src/com/android/launcher3/dagger/LauncherComponentProvider.kt#L35

### Testing

If you want to see this issue you can install or uninstall an app and observe that the app list is never updated.

### Type of change

## Summary by CodeRabbit

* **Chores**
  * Adjusted application startup/integration to change how the launcher initializes during app creation.
  * Modified provider creation behavior so the previous secondary initialization step is skipped, affecting the startup sequence.

* **Bug Fixes**
  * Minor startup stability adjustments expected; no user-facing UI changes.

### DIFF
--- a/lawnchair/src/app/lawnchair/LawnchairApp.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairApp.kt
@@ -18,7 +18,6 @@ package app.lawnchair
 
 import android.annotation.SuppressLint
 import android.app.Activity
-import com.android.launcher3.LauncherApplication
 import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
@@ -50,6 +49,7 @@ import app.lawnchair.views.ComposeBottomSheet
 import com.android.launcher3.BuildConfig
 import com.android.launcher3.InvariantDeviceProfile
 import com.android.launcher3.Launcher
+import com.android.launcher3.LauncherApplication
 import com.android.launcher3.R
 import com.android.launcher3.Utilities
 import com.android.quickstep.RecentsActivity
--- a/lawnchair/src/app/lawnchair/LawnchairApp.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairApp.kt
@@ -18,7 +18,7 @@ package app.lawnchair
 
 import android.annotation.SuppressLint
 import android.app.Activity
-import android.app.Application
+import com.android.launcher3.LauncherApplication
 import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
@@ -56,7 +56,7 @@ import com.android.quickstep.RecentsActivity
 import com.android.systemui.shared.system.QuickStepContract
 import java.io.File
 
-class LawnchairApp : Application() {
+class LawnchairApp : LauncherApplication() {
     private val compatible = Build.VERSION.SDK_INT in BuildConfig.QUICKSTEP_MIN_SDK..BuildConfig.QUICKSTEP_MAX_SDK
     private val isRecentsComponent: Boolean by unsafeLazy { checkRecentsComponent() }
     private val recentsEnabled: Boolean get() = compatible && isRecentsComponent
--- a/src/com/android/launcher3/LauncherProvider.java
+++ b/src/com/android/launcher3/LauncherProvider.java
@@ -119,7 +119,6 @@ public class LauncherProvider extends ContentProvider {
 
     @Override
     public boolean onCreate() {
-        MainProcessInitializer.initialize(getContext().getApplicationContext());
         return true;
     }
 


strangr:
just a question here,

Why doesn't `LawnchairApp` extend `LauncherApplication`? 
if it did, this problem would not have existed as 

`LauncherComponentProvider` would have stopped at 

`if (app is LauncherApplication) return app.appComponent`


Im just bringing this up to the attention, as I do not know true answer to this, but maybe who does will notice bigger similar problems

Mattwmaster58:
@strangr I'm not sure the exact reason why it inherits only from Application, but I do know that 15 did it the same way. And actually, launcher app component is pretty thin, so inheriting from there seems like it is the "right" way to actually do this, I think you're right.

SuperDragonXD:
> I'm not sure the exact reason why it inherits only from Application, but I do know that 15 did it the same way.

To clarify this, `15-dev` didn't inherit from `LauncherApplication`, because in previous versions (particularly before `15-dev`), that said component didn't exist at all.

Then, it was in Launcher3 from A15 where [they re-implemented this class back](79085dafd4431ea4e0b29ea1bc0fe81807d65629), then Launcher3 from A16 is when they fully migrated parts of the app to use Dagger. And it seems like we haven't updated `LawnchairApp` to use it yet.

Will still need to wait for a review from @validcube to confirm if this is the right fix.

validcube (approved):
LGTM.

> I don't have a good answer for why this was never done before and know the FULL implications of such a change. It seems like it was somewhat intentional, this comment would have been irrelevant (we would never hit the paths the engage safe mode) IF this approach was taken from the start, so clearly someone was aware of the approach being taken:

Back then around June 2025, all temporary code changes to workaround Launcher3 new requirements have a comment starting with `Lawnchair-Notice` then by around Sep 2025 all the workaround starting with `Lawnchair-Notice` have been replaced actual fixes but this one slipped past my review so, nice catch to all of you.

I'll take it from here and review if there are anymore of these old comment to fix.

PR #6636, opened 2026-04-14
fix: Link app drawer background opacity to correct colors

validcube:
### Description

Make allapps transparency work again, and calculate the best visual settings for blurred/non-blurred state.

### Reasoning

Link the opacity colour to the right colour constant, and make sure that the opacity carry-over to two separate colour for blurred and non-blurred. Before we (or I) linked it to the scrim background instead of the foreground of the widget picker.

Additionally: Opacity now has two specs for blurred and non-blurred state so that opacity settings designed for blurred would look nicer on those designed for non-blurred. And alternate text threshold has been changed from simple opacity >= .3f to WCAG 2.0 luminance > .5f for much better calculation.

### Testing

Turn on and off window blur feature then change opacity and swipe up app drawer.

### Type of change

✅ **Bug fix** (A non-breaking change that fixes an issue)

## Summary by CodeRabbit

* **Refactor**
  * Unified all‑apps and scrim color handling to consistently respect drawer background overrides, derive text color from base luminance plus opacity for improved contrast, and use shared color tokens for cleaner visuals.
  * Header scrim blending now computes alpha from visible UI transparency and blends against the current background for more accurate appearance.

* **Performance**
  * Cached bottom‑sheet background and lifecycle‑aware blur listeners to reduce unnecessary redraws and keep visuals synchronized.


lawnchair/src/app/lawnchair/util/LawnchairUtils.kt
```diff
@@ -180,17 +186,31 @@ fun getFolderBackgroundAlpha(context: Context): Int {
...
+    val prefs = PreferenceManager.getInstance(context)
+    val isBlurred = ActivityContext.lookupContext<BaseActivity>(context).isAllAppsBackgroundBlurEnabled
+    val userOpacity = prefs.drawerOpacity.get()
+
+    val actualOpacity = calculateAllAppsBestVisualOpacity(userOpacity, isBlurred)
+    return ColorUtils.setAlphaComponent(getAllAppsBaseColor(context, defaultColor), (actualOpacity * 255).roundToInt())
+}
+
+private fun calculateAllAppsBestVisualOpacity(alpha: Float, isBlurred: Boolean): Float {
+    val allAppsBlurRange = 0.3f..1.0f
```
SuperDragonXD: If I understand this, you can no longer make the app drawer background transparent when blur is on?

Can you send a screenshot for this to check?

validcube: Yeah, but now that I think about it this is not really needed anymore since it's succeeded by a much better check in overrideAllAppsTextColor. I'll revert this.

SuperDragonXD (approved):

PR #6745, opened 2026-05-13
feat: Reimplement support for non-root app predictions

### LINKED ISSUE #4235: [FEATURE] Allow viewing recently or frequently used apps in app drawer
### Describe the feature

Ability to see recently used apps at the top of the app drawer.

My most used apps will be at the top of the app drawer.

No

### Example(s)

google launcher has this feature.

### Additional context

_No response_

### LINKED ISSUE #6426: [FEATURE] Suggested app On Dock
### Describe the feature

As stated in the tile it would be very useful to have a suggested apps thing in the dock

I find it quicket to access apps

### Did the feature exist in Lawnchair Legacy?

No

### Example(s)

Image

### Additional context

_No response_

validcube:
### Description

Lawnchair rootless app prediction system. (Basic app ranking... Basic?)

Fixes #4235

Todo:
- [x] App Predictor that closely resembles one from `QuickstepModelDelegate`
  - [x] Use usage stats (not system api) on first data when possible 
  - [x] Combine usage stats signal (everyday signal) with predictor
  - [x] React to on suggestion discard
  - [x] ~~Headphone store for hotseat + allapps~~ No longer planned for this PR since that would likely be too much work.
- [x] Promote to stable (optional in this PR)
  - [ ] Design/Discuss passable UI/UX
    - [x] Allow user to view all suggested apps in the store (and mark those apps as dismissed/do not appear)
  - [x] Allow switching prediction modal
  - [x] ~~Allow changing prediction "influencer"~~ Headphone store and other device state store is supposed to be the "influencer" but now that it's no longer planned, this will not be implemented.

#### Code side

~~Heads up note: All of the files were written before AOSP Proto files synced with the IDE so there may be some weird usage around `LauncherAtom` as I use AI with pretty much "Claude, integrate `LauncherAtom` with my code, make no mistake (the IDE is lying to you about missing files)"~~ All files were cleaned up now.

This changes override predictor from `QuickstepModelDelegate` when predictor mode is set to Lawnchair instead of System.

The Lawnchair predictor a basic ranking system based on how many time the app was opened, it tried to be very accurate to how the system predictor returns it datas (because it has to be accepted by the AOSP, thanks AI for doing that painful tedious work) and works by listening in on app launch and event stats returned by the `StatsLogs`.

If a data from a store is not available like hotseat is non existent it will use random data (ranked lowest) and data from other store like allapps (ranked higher than random).

Compared to old prediction system in used by Lawnchair 10 this one does not take wired/wireless headphone into context. And since most of the changes there are not salavagable/technically infeasible we would have to write one from scratch.

#### UI side

The feature will be part of Prediction which will house all future Lawnchair prediction feature there, I envision it to also include feature like folder naming suggestions.

image

The switch act as global toggle in case some people don't like prediction features and wish to turn it off in one go.

https://github.com/user-attachments/assets/9e4b731b-4391-4f14-842a-4b711b114992

The main switch on the top of the page will also act as global toggle.

> [!NOTE]
> The animation will NOT be included in this PR, to restore this change revert bd5a06b793cce5709c9eaa26276552582f9b4c4c, bd5a06b793cce5709c9eaa26276552582f9b4c4c, then 81d04387eceed3466178b415b565a5c6a29b4f71

<details><summary>Cut from this PR // Animation rationale</summary>
<p>

##### UI side // Animation rationale

While animation will not be included in the PR, in case it is ever needed in the future.

The animation is originally designed with Lottie but there are some limitations that I personally feels like Android Animated Vector Drawable (AVD) is easier so it was ported to AVD. I designed in a way that it can be easily repeated or trick user that it's respond to preference signal by looping the animation when the feature is on, or stop looping when the feature is off.

Unfortunately there are some downside with the animation is that:
1. You can't programatically do dynamic theming

</p>
</details> 

### Testing

Install the latest build from this PR or build locally, on first use the predictor will randomly generate a list of suggested apps to use. which you can see when you open the allapps, hotseat, and widget selector.

The datastore in allapps and hotseat are separate meaning allapps will only show prediction for app that you open from allapps, and hotseat show prediction for app you open from hotseat.

However in an event where the user didn't have data on one of the store like say user don't have on allapps, the predictor will fallback to grabbing data from other store if exist then fallback to random data if data didn't exist at all but they will always be ranked lower than what's in allapps store.

When you have usage stats turn on for Lawnchair, instead of grabbing random data, Lawnchair will grab the best apps user is likely to open from usage stats instead.

### Type of change

✅ **New feature** (A non-breaking change that adds functionality)

## Summary by CodeRabbit

- **New Features**
  - Added experimental App suggestions settings with System, Lawnchair, or Off modes.
  - Added controls for usage-based ranking, prediction enablement, and dismissed apps.
  - Added a screen to review, restore, or reset dismissed app suggestions.
  - Lawnchair predictions now learn from app launches and provide app, hotseat, and widget suggestions.
- **Bug Fixes**
  - Improved restoration and refreshing of predictions when returning to the launcher or completing transitions.
  - Added guidance when usage-access permission is required for usage-based suggestions.


lawnchair/src/app/lawnchair/predictions/LawnchairAppPredictor.kt
```diff
@@ -0,0 +1,618 @@
...
+            }
+
+            resolvedEvent.location.startsWith("all-apps") ||
+                resolvedEvent.location == "predictions" -> {
+                allAppsStore.record(key)
+                true
+            }
+
+            else -> false
+        }
```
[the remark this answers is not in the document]
validcube: It used to be recorded in prior to today commit however this may create some sort of feedback loop where you will be freezing the prediction as you will keep ranking already predicted app higher than other candidates. Personally I think removing it is the best decisions since the current behaviour already does the job while still providing relevant result unless you think otherwise and think that there are people who would prefer stable/stills predictions.

lawnchair/src/app/lawnchair/predictions/PredictionMode.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+object SystemPredictor : PredictionMode(R.string.prediction_mode_system) {
+
+    override fun toString() = "system"
+
+    override fun isAvailable(context: Context): Boolean {
+        if (context.getSystemService(AppPredictionManager::class.java) == null) return false
+        return context.checkSelfPermission(Manifest.permission.PACKAGE_USAGE_STATS) ==
+            PackageManager.PERMISSION_GRANTED // It needs to be granted through ADB only, not through special app appops.
```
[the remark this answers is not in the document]
validcube: This is unfortunately intended as the AppPrediction API (system predictor) provided by Android enforce package usage stats as runtime permissions instead of AppOpps likely because it was a @systemapi and they probably didn't see the need to allow via AppOpps.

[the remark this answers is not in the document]
validcube: @coderabbitai

lawnchair/res/values/strings.xml
```diff
@@ -201,6 +201,10 @@
     <string name="app_info_drop_target_label">App info</string>
     <string name="debug_restart_launcher">Restart Lawnchair</string>
     <string name="experimental_features_label">Experimental features</string>
+    <string name="predictions_label">Predictions</string>
+    <string name="global_predictions_label">Prediction Features</string>
```
SuperDragonXD:
```suggestion
    <string name="global_predictions_label">Show prediction features</string>
```

lawnchair/src/app/lawnchair/ui/preferences/destinations/ExperimentalFeaturesPreferences.kt
SuperDragonXD, line 1: This should be a separate PR/direct commit.

lawnchair/src/app/lawnchair/ui/preferences/destinations/GeneralPreferences.kt
```diff
@@ -144,6 +147,18 @@ fun GeneralPreferences() {
...
+        PreferenceGroup {
+            Item {
+                NavigationSwitchPreference(
+                    checked = enableGlobalPrediction.state.value,
+                    onCheckedChange = enableGlobalPrediction::onChange,
+                    label = stringResource(R.string.predictions_label),
+                    description = stringResource(R.string.predictions_description), // In future this can be extended to include folder naming suggestions.
+                    onClick = { navController.navigate(Predictions) },
+                )
+            }
```
SuperDragonXD: We can simplify this by just making it a simple `NavigationPreference`. Tapping directly should just visit the page immediately, similar to the Pixel Launcher.

lawnchair/res/values/strings.xml
```diff
@@ -209,6 +213,16 @@
...
 
+    <string name="prediction_mode_label">Prediction mode</string>
+    <string name="prediction_mode_system">System</string>
+    <string name="prediction_mode_lawnchair">Lawnchair</string>
+    <string name="prediction_mode_none">None</string>
+    <string name="prediction_weighted_usage_stats_label">Weighted usage stats</string>
+    <string name="prediction_weighted_usage_stats_description">Use usage stats to influence ranking suggestions</string>
+    <string name="prediction_weighted_usage_stats_permission_description">Use usage stats to influence ranking suggestions. Requires usage access to enable</string>
+    <string name="dismissed_prediction_apps_label">Dismissed prediction apps</string>
+    <string name="dismissed_prediction_apps_label_with_count">Dismissed prediction apps (%1$d)</string>
```
SuperDragonXD:
```suggestion
    <string name="dismissed_prediction_apps_label">Blocked apps</string>
    <string name="dismissed_prediction_apps_label_with_count">Blocked apps (%1$d)</string>
```

lawnchair/src/app/lawnchair/ui/preferences/destinations/AppDrawerPreferences.kt
```diff
@@ -88,7 +87,6 @@ fun AppDrawerPreferences(
                 )
             }
             Item { SearchBarPreference(SearchRoute.DRAWER_SEARCH, showLabel = false) }
-            SuggestionsPreference()
```
SuperDragonXD, line 92: We should still keep the preference here, to preserve existing behavior.

lawnchair/src/app/lawnchair/ui/preferences/components/SystemSuggestionsPreference.kt
SuperDragonXD, line 1: Move this to `PredictionsPreferences` to centralize all suggestions-preference state there.

lawnchair/src/app/lawnchair/ui/preferences/destinations/DismissedPredictionAppsPreferences.kt
```diff
@@ -0,0 +1,172 @@
...
+    }
+    var dismissedApps by remember {
+        mutableStateOf(dismissedAppsStore.getDismissedApps())
+    }
+    val pageTitle =
+        if (dismissedApps.isEmpty()) {
+            stringResource(id = R.string.dismissed_prediction_apps_label)
+        } else {
+            stringResource(id = R.string.dismissed_prediction_apps_label_with_count, dismissedApps.size)
+        }
```
SuperDragonXD: Wrap this inside a `stringResource` block to reduce repetition:


```suggestion
    val pageTitle = stringResource(
        if (dismissedApps.isEmpty()) {
            R.string.dismissed_prediction_apps_label
        } else {
             R.string.dismissed_prediction_apps_label_with_count, dismissedApps.size
        }
    )
```

Also, I think we can inline this at PreferenceScaffold.

lawnchair/src/app/lawnchair/predictions/DismissedPredictionAppsStore.kt
SuperDragonXD, line 1: See above file.

lawnchair/src/app/lawnchair/predictions/AppUsageStore.kt
```diff
@@ -0,0 +1,63 @@
+package app.lawnchair.predictions
+
+import android.content.Context
+import android.content.SharedPreferences
+import android.content.pm.PackageManager
+import androidx.core.content.edit
+import app.lawnchair.util.isPackageInstalled
+
+class AppUsageStore(
+    private val prefs: SharedPreferences,
```
SuperDragonXD: I'd suggest storing this in Preference DataStore and abstracting it via `PreferenceManager`. This way, we can just directly store the lists without writing converters, while also matching the internal docs we have about preferring the newer preference standard.

See [`LiveInformationManager`](https://github.com/LawnchairLauncher/lawnchair/blob/16-dev/lawnchair/src/app/lawnchair/ui/preferences/data/liveinfo/LiveInformationManager.kt) for an example.

lawnchair/src/app/lawnchair/ui/preferences/components/controls/NavigationSwitchPreference.kt
SuperDragonXD, line 1: We won't need this per my comment at `GeneralPreferences`.

build.gradle
SuperDragonXD, line 1: Is this relevant for the PR or no?

lawnchair/src/app/lawnchair/predictions/AppUsageStore.kt
SuperDragonXD, line 1: We don't need to have two distinct data storage solutions here. I think we can consolidate both classes into something like this:

```kotlin
class LawnchairPredictionStore(
    private val isOrdered: Boolean = false,
) {
    // Either store cache as a mutable list of string OR a string set, not both
    private var cache: MutableList<String> = load()
     
    // replaces record()
    fun add(key: String) {
        if (key.isEmpty()) return
        if (isOrdered) {
            // for the default
            cache.add(0, key)
            if (cache.size > maxSize) {
                cache = cache.take(maxSize).toMutableList()
            }
        } else {
            // for dismissed
            if (cache.contains(key)) return
            cache.add(key)
        }
        save()
    }
    fun remove(key: String)
    fun getEntries()
    fun getRanked()
    fun pruneUninstalled(pm: PackageManager)
   
    // Implement load() and save() by serializing/deserializing the string.
}
```

Also, please use idiomatic Kotlin, like `.filter { it.isNotEmpty }` instead of `.filter(String::isNotEmpty)` (as the former is more common).

lawnchair/src/app/lawnchair/predictions/LawnchairAppPredictor.kt
SuperDragonXD, line 1: Please separate this class into multiple classes for maintainability. I've given boilerplate code as a start:

`PredictionEventResolver`:

```kotlin
class PredictionEventResolver(private val userCache: UserCache) {
    data class ResolvedEvent()
    
    fun resolve(atomInfo: ItemInfo?): ResolvedEvent? {
        if (atomInfo == null) return null
        return ResolvedEvent(
            location = resolveLocation(atomInfo),
            componentName = resolveComponentName(atomInfo),
            user = resolveUser(atomInfo)
        )
    }

    private fun resolveComponentName(atomInfo: ItemInfo): ComponentName?
    private fun resolveUser(atomInfo: ItemInfo): UserHandle
    private fun resolveLocation(atomInfo: ItemInfo): String
```

----

`UsageStatsRanker`:

```kotlin
class UsageStatsRanker(private val context: Context) {

    fun getUsageStatsRanked(
        launcherApps: LauncherApps,
        appFilter: AppFilter,
        resolveStoreKey: (String, LauncherApps, AppFilter) -> String?
    ): List<String> {
    // list all windows, then do .forEach calling addUsageStatsToScores
    }

   private fun addUsageStatsToScores(
        scores: MutableMap<String, Double>,
        stats: Map<String, UsageStats>,
        now: Long,
        window: UsageStatsWindow
    )

    private data class UsageStatsWindow(
        val durationMs: Long,
        val launchWeight: Double,
        val foregroundMinutesWeight: Double,
        val InnovationsWeight: Double = 0.0,
        val recencyWeight: Double
    )
    
    // companion object of all the usage stats constants
```

`LawnchairPredictionEngine`:

```kotlin
class LawnchairPredictionEngine(
    private val context: Context,
    private val usageStatsRanker: UsageStatsRanker
) {
    fun compileAppTargets(...): List<AppTarget>
    fun compileWidgetTargets(...): List<AppTarget>
    fun getFallbackRanked(...)
    fun mergeRanked(...)
    
    private fun getRandomRanked()
    private fun createAppTarget(...)

   // hold everything that doesn't need the system state
}
```

Which makes `LawnchairAppPredictor` get all the above information and updates the system, etc.

lawnchair/src/app/lawnchair/ui/preferences/destinations/DismissedPredictionAppsPreferences.kt
SuperDragonXD, line 1: This file will be changed heavily once you implemented my suggestions regarding AppUsageStore,

lawnchair/res/values/strings.xml
```diff
@@ -209,6 +213,16 @@
     <string name="internal_label">Internal</string>
     <string name="internal_description">Internal may be more unstable than other experimental group</string>
 
+    <string name="prediction_mode_label">Prediction mode</string>
+    <string name="prediction_mode_system">System</string>
+    <string name="prediction_mode_lawnchair">Lawnchair</string>
+    <string name="prediction_mode_none">None</string>
+    <string name="prediction_weighted_usage_stats_label">Weighted usage stats</string>
+    <string name="prediction_weighted_usage_stats_description">Use usage stats to influence ranking suggestions</string>
+    <string name="prediction_weighted_usage_stats_permission_description">Use usage stats to influence ranking suggestions. Requires usage access to enable</string>
```
SuperDragonXD:
```suggestion
    <string name="prediction_weighted_usage_stats_label">Use system predictions</string>
    <string name="prediction_weighted_usage_stats_description">Learns from your system usage habits to provide better app suggestions</string>
    <string name="prediction_weighted_usage_stats_permission_description">Use usage stats to influence ranking suggestions. Requires usage access to enable</string>
```

lawnchair/src/app/lawnchair/ui/preferences/destinations/PredictionsPreferences.kt
```diff
@@ -0,0 +1,178 @@
...
+    hasUsageStatsPermission: Boolean,
+    dismissedPredictionAppsSubtitle: String,
+) {
+    Item {
+        SwitchPreference(
+            adapter = weightedUsageStatsAdapter,
+            label = stringResource(R.string.prediction_weighted_usage_stats_label),
+            description = weightedUsageStatsDescription,
+            enabled = hasUsageStatsPermission,
+        )
```
SuperDragonXD: Shouldn't we guard this with a permissiom check (and prompt the user to enable usage statistics)?

lawnchair/src/app/lawnchair/preferences2/PreferenceManager2.kt
```diff
@@ -372,6 +373,23 @@ class PreferenceManager2 @Inject constructor(
...
+        key = stringPreferencesKey(name = "prediction_mode"),
+        defaultValue = PredictionMode.fromString(context.getString(R.string.config_default_prediction_mode)),
+        parse = { PredictionMode.fromString(it) },
+        save = { it.toString() },
+    )
+
+    val lawnchairPredictorUseWeightedUsageStats = preference(
+        key = booleanPreferencesKey(name = "lawnchairpredictor_use_weighted_usage_stats"),
+        defaultValue = false,
+    )
```
SuperDragonXD, line 403: Use a simpler key instead, like `lawnchair_prediction_use_weighted_usage`.

lawnchair/res/values/strings.xml
```diff
@@ -201,6 +201,10 @@
     <string name="app_info_drop_target_label">App info</string>
     <string name="debug_restart_launcher">Restart Lawnchair</string>
     <string name="experimental_features_label">Experimental features</string>
+    <string name="predictions_label">Predictions</string>
+    <string name="global_predictions_label">Prediction Features</string>
+    <string name="predictions_description">App predictions</string>
+    <string name="app_predictions_label">App predictions</string>
```
SuperDragonXD: We don't need two strings that have the same contents.

lawnchair/res/values/strings.xml
```diff
@@ -209,6 +213,16 @@
     <string name="internal_label">Internal</string>
     <string name="internal_description">Internal may be more unstable than other experimental group</string>
 
+    <string name="prediction_mode_label">Prediction mode</string>
```
SuperDragonXD:
```suggestion
    <string name="prediction_mode_label">Predictions source</string>
```

lawnchair/res/values/strings.xml
```diff
@@ -209,6 +213,16 @@
...
+    <string name="prediction_mode_system">System</string>
+    <string name="prediction_mode_lawnchair">Lawnchair</string>
+    <string name="prediction_mode_none">None</string>
```
SuperDragonXD: Use the `disabled` (preferably `off`) and `system` strings instead.

lawnchair/src/app/lawnchair/predictions/PredictionAppKey.kt
```diff
@@ -0,0 +1,33 @@
...
+        val userTokenStart = key.lastIndexOf(DELIMITER)
+        if (packageNameEnd <= 0 || userTokenStart <= packageNameEnd + 1 || userTokenStart >= key.lastIndex) {
+            return null
+        }
+
+        return TargetInfo(
+            packageName = key.substring(0, packageNameEnd),
+            className = key.substring(packageNameEnd + 1, userTokenStart),
+            userToken = key.substring(userTokenStart + 1),
+        )
```
SuperDragonXD:
```suggestion
        val parts = key.split(DELIMITER)
        if (parts.size != 3 || parts.any { it.isEmpty() }) return null
        return TargetInfo(parts[0], parts[1], parts[2])
```

lawnchair/src/app/lawnchair/ui/preferences/destinations/PredictionsPreferences.kt
```diff
@@ -0,0 +1,178 @@
...
+            if (key == DismissedPredictionAppsStore.DISMISS_STORE_NAME) {
+                dismissedPredictionAppsCount = dismissedAppsStore.getDismissedApps().size
+            }
+        }
+        predictionPrefs.registerOnSharedPreferenceChangeListener(listener)
+        onDispose { predictionPrefs.unregisterOnSharedPreferenceChangeListener(listener) }
+    }
+
+    return dismissedPredictionAppsCount
+}
```
SuperDragonXD, line 205: This will be changed once my suggestions are implemented.

lawnchair/src/app/lawnchair/predictions/PredictionAppKey.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    data class TargetInfo(
+        val packageName: String,
+        val className: String,
+        val userToken: String,
+    )
+
+    fun create(componentName: ComponentName, userToken: String): String = "${componentName.packageName}$DELIMITER${componentName.className}$DELIMITER$userToken"
+
+    fun create(componentName: ComponentName, userSerial: Long): String = create(componentName, userSerial.toString())
```
SuperDragonXD: For consistency, wouldn't it better to store this as only `userSerial` or `userHandle` only?

lawnchair/src/app/lawnchair/ui/preferences/destinations/DismissedPredictionAppsPreferences.kt
```diff
@@ -0,0 +1,172 @@
...
+    val apps by appsState(comparator = dismissedPredictionAppsComparator(context, dismissedApps))
+    val state = rememberLazyListState()
```
SuperDragonXD: We can inline this at `PreferenceLazyColumn`.

SuperDragonXD (changes requested):
I have several suggestions in improving this feature, particularly in the `predictions` package. I'll comment on the UI/UX properly later on.

build.gradle
validcube: Yes, the predictor uses `LauncherAtom.proto` from AOSP and the changes here made to fix IDE unable to found proto files from AOSP.

But I don't think this is the perfect solution though.

lawnchair/res/values/strings.xml
validcube: I think it would be better to use "Dismiss" wording for better consistency with what the AOSP shows to user with prediction enabled.

SuperDragonXD: Yeah, that can work. So we can do this instead:

```suggestion
    <string name="dismissed_prediction_apps_label">Dismissed apps</string>
    <string name="dismissed_prediction_apps_label_with_count">Dismissed apps (%1$d)</string>
```

lawnchair/res/values/strings.xml
validcube: I realised that this will conflict with Android system predictor mode (`AppPrediction` API), I suggest we call it something else?

lawnchair/src/app/lawnchair/ui/preferences/destinations/DismissedPredictionAppsPreferences.kt
validcube: I tested it and I would have to revert it because Kotlin if only allow a single object to return which is the value that will be pass to the stringResource first args.

lawnchair/src/app/lawnchair/ui/preferences/destinations/GeneralPreferences.kt
validcube: The switch navigation is intentional, the navigation serve as a future idea for where to put prediction-based features and the switch preference to toggle the prediction-based features globally in one switch.

If you want to make it house only the app predictions then sure we can do that.

lawnchair/src/app/lawnchair/ui/preferences/destinations/PredictionsPreferences.kt
validcube: I just realised that the app usage stats feature currently doesn't do anything substantial yet other than just show relevant app on first use only and nothing more. It would be better to remove or hide the toggle for now since it would likely not affect any user except third-party custom ROM solution that offer Lawnchair with fullest permissions as the default launcher.

What do you think?

Alternatively we could try to implement this like usage + routine-based prediction but I think that would requires a lot of fine-tuning although the code is there, just underutilised.

lawnchair/src/app/lawnchair/ui/preferences/destinations/GeneralPreferences.kt
```diff
@@ -59,6 +60,7 @@ fun GeneralPreferences() {
     val context = LocalContext.current
     val prefs = preferenceManager()
     val prefs2 = preferenceManager2()
+    val navController = LocalNavController.current
```
SuperDragonXD: Just remove this unused import.

lawnchair/src/app/lawnchair/ui/preferences/destinations/PredictionsPreferences.kt
validcube: Nevermind, should be addressed in the latest commit.

SuperDragonXD (approved):

PR #6897, opened 2026-06-18
perf: Cache PreferenceManager2 in memory

### LINKED ISSUE #6878: Cache preference in memory
### Describe the feature

Cache preference key and value in memory, avoiding `.firstBlocking()` operandi

Faster access to preferences, avoid having us to cache response manually when a function fired the preference `.firstBlocking()` repeatedly

### Did the feature exist in Lawnchair Legacy?

No

### Example(s)

1. Store preference in memory
2. Update preference in memory on preference changed

### Additional context

_No response_

validcube:
### Description

Fixes #6878 

### Reasoning

Cache instead of run blocking every time prefs are accessed, changes applied to preference will be immediately synced everywhere.

For compatibility reason, this will not remove `.firstBlocking()` but instead deprecate it so that it won't be a breaking changes for existing PR, and allows more time for other user to adapt to the new `.firstCached(prefs2)`.

### Testing

Prefs test

### Type of change

✅ **Performance** (A code change that improves performance)

## Summary by CodeRabbit

* **Performance**
  * Improved responsiveness by switching many in-app preference reads to cached, non-blocking values across launcher UI, search, theming, gestures, and workspace/taskbar behaviors.
  * Added an in-memory preference cache to keep interactive updates feeling immediate.
* **Stability**
  * More consistent feature gating and rendering (including Smartspace and search) by relying on continuously refreshed cached preferences.
  * Properly stops background preference caching work when closing.
* **Developer Experience**
  * Added cached preference utilities (including lifecycle-aware Compose support) and discouraged blocking preference reads.

### DIFF
--- a/lawnchair/src/app/lawnchair/DeviceProfileOverrides.kt
+++ b/lawnchair/src/app/lawnchair/DeviceProfileOverrides.kt
@@ -96,20 +96,20 @@ class DeviceProfileOverrides @Inject constructor(
             prefs2: PreferenceManager2,
             defaultGrid: InvariantDeviceProfile.GridOption,
         ) : this(
-            numAllAppsColumns = prefs2.drawerColumns.firstCached(gridOption = defaultGrid, prefs2 = prefs2),
+            numAllAppsColumns = prefs2.drawerColumns.firstCached(gridOption = defaultGrid),
             numFolderRows = prefs.folderRows.get(defaultGrid),
-            numFolderColumns = prefs2.folderColumns.firstCached(gridOption = defaultGrid, prefs2 = prefs2),
+            numFolderColumns = prefs2.folderColumns.firstCached(gridOption = defaultGrid),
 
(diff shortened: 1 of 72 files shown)


SuperDragonXD (commented):
Is there a way to make the call sites less clunky, perhaps by using [context parameters](https://kotlinlang.org/docs/context-parameters.html) for `PreferenceImpl.firstCached()`?

If that's not possible as of now, we can just merge as is.

validcube:
> Is there a way to make the call sites less clunky, perhaps by using [context parameters](https://kotlinlang.org/docs/context-parameters.html) for `PreferenceImpl.firstCached()`?
> 
> If that's not possible as of now, we can just merge as is.

I have simplified the code although by providing a default prefs2 to the `firstCached` parameter instead of using context parameter since I figured that it would be less verbose.

SuperDragonXD (approved):

PR #6899, opened 2026-06-18
refactor: Canonically accurate bottomsheet

validcube:
### Description

Convert Lawnchair bottom sheet use with Launcher3 code to actual androidx buttom sheet.

This allow us to use expressive physics everywhere and make bottom sheet partial on area that needs it like CustomizeDialog

### Testing

Manual trigger

### Type of change

✅ **Refactor** (A code change that neither fixes a bug nor adds a feature)

## Summary by CodeRabbit

* **New Features**
  * Updated dialogs, prompts, and customization flows with a more consistent bottom-sheet experience.
  * Improved bottom-sheet interactions across settings, shortcuts, About, and icon-pack application flows.

* **Bug Fixes**
  * Improved dismissal behavior for confirmation, warning, and incompatibility prompts.
  * Refined gesture, sleep, accessibility, and App info actions so sheets close more reliably.


lawnchair/src/app/lawnchair/override/CustomizeDialog.kt
```diff
@@ -168,43 +182,67 @@ fun CustomizeAppDialog(
...
+        animationSpec = spring(stiffness = Spring.StiffnessMediumLow),
+        label = "BottomSheetBlurFraction",
+    )
+    val scrimAlpha = .32f * animatedFraction
+
+    ModalBottomSheet(
+        onDismissRequest = onClose,
+        sheetState = bottomSheetState,
+        containerColor = MaterialTheme.colorScheme.surface,
+        scrimColor = MaterialTheme.colorScheme.onSurface.copy(alpha = scrimAlpha),
```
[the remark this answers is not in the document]
validcube: This is intended behaviour, without darkening it would be distracting since the contrast is usually not in the bottom sheet favour.

lawnchair/src/app/lawnchair/ui/util/ProvideBottomSheetHandler.kt
```diff
@@ -132,25 +135,16 @@ fun ProvideBottomSheetHandler(
...
-                        showBottomSheet = false
-                    },
-                    contentWindowInsets = {
-                        windowInsets
-                    },
-                    scrimColor = if (supportsBlur) Color.Transparent else BottomSheetDefaults.ScrimColor,
-                ) {
-                    bottomSheetContent.content()
-                }
+                bottomSheetContent.content()
```
SuperDragonXD: Is there any advantage to moving the bottom sheet code from here to `AlertBottomSheetContent`? It feels like additional boilerplate.

lawnchair/src/app/lawnchair/views/ComposeBottomSheet.kt
```diff
@@ -1,62 +1,45 @@
...
 import app.lawnchair.util.ProvideLifecycleState
-import app.lawnchair.util.minus
 import com.android.launcher3.Launcher
-import com.android.launcher3.anim.PendingAnimation
-import com.android.launcher3.util.SystemUiController
 import com.android.launcher3.views.AbstractSlideInView
 import com.android.launcher3.views.ActivityContext
 import com.android.launcher3.views.BaseDragLayer
 
 class ComposeBottomSheet<T>(context: Context) : AbstractSlideInView<T>(context, null, 0) where T : Context, T : ActivityContext {
```
SuperDragonXD: If we migrate to a Compose-backed view, do we necessarily have to extend `AbstractSlideInView`? Considering that we don't really use the methods here anymore.

lawnchair/src/app/lawnchair/gestures/handlers/GestureWithAccessibilityHandler.kt
```diff
@@ -22,38 +27,61 @@ import app.lawnchair.lawnchairApp
...
+                    initialValue = SheetValue.Hidden,
+                    enabledValues = setOf(SheetValue.Hidden, SheetValue.Expanded),
+                )
+                val coroutineScope = rememberCoroutineScope()
+                val closeSheet = {
+                    coroutineScope.launch { sheetState.hide() }.invokeOnCompletion {
+                        close(true)
+                    }
+                    Unit
+                }
```
SuperDragonXD: We can inline this calls to a new function `rememberSheetState()` found in `ProvideBottomSheetHandler` to reduce boilerplate.

Same with `closeSheet` as well.

SuperDragonXD (changes requested):
Hello. Just have a few comments for this PR.

lawnchair/src/app/lawnchair/views/ComposeBottomSheet.kt
validcube: We still do, although less of it now since everything has moved to compose. We need it to so that the compose layer is attached with the AOSP base drag layer.

Since we don't use the entire `AbstractSlideInView`, I think we could migrate to `AbstractFloatingView` which is a leaner alternative.

lawnchair/src/app/lawnchair/ui/util/ProvideBottomSheetHandler.kt
validcube: Should be addressed in latest commit.

lawnchair/src/app/lawnchair/views/ComposeBottomSheet.kt
validcube: Should be addressed in latest commit.

lawnchair/src/app/lawnchair/gestures/handlers/GestureWithAccessibilityHandler.kt
validcube: Should be addressed in latest commit.

PR #6950, opened 2026-07-03
Scale the maximum home screen grid rows with available screen height

Broikos-Nikos:
## What

The home screen grid **rows** slider was capped at a flat **10** (20 with the
experimental "increase max grid size" toggle). On tall/large devices there's room
for more, so this makes the maximum scale with the device instead of being fixed.

## How

`maxRows` is derived from **density-independent pixels (dp)**, not raw pixels:

```
dp      = longerScreenEdgePx / displayMetrics.density
maxRows = (dp / ~60dp-per-row) coerced into 10..20
```

Only the **rows** slider grows — columns and dock icons keep the existing limit,
and the experimental toggle still unlocks the full range on any device. This only
raises the *upper bound* the user can select; the default grid is unchanged and
cells scale to fit whatever is chosen.

## Why dp instead of pixels

Android lays out UI in dp, and the same pixel count maps to different physical
sizes across densities — so a raw-pixel threshold would hand the most rows to the
devices with the *least* usable room. For example a low-density 720×1600 phone has
~1067 dp of height, while a high-density 1080×2400 phone has only ~800 dp. AOSP
Launcher3 itself converts to dp (`dpiFromPx`) before selecting a grid, so this
follows the platform's own approach.

Resulting max rows on representative devices:

| Device | Resolution / density | dp height | Max rows |
|---|---|---|---|
| Budget HD+ (low density) | 720×1600 @1.5× | 1067 | 17 |
| Mainstream FHD+ | 1080×2400 @3.0× | 800 | 13 |
| Flagship QHD+ | 1440×3200 @3.5× | 914 | 15 |
| Pixel 9 Pro XL | 1344×2992 @2.81× | 1064 | 17 |
| Tablet | 1600×2560 @2.0× | 1280 | 20 (cap) |

## Files changed

- `lawnchair/src/app/lawnchair/ui/preferences/destinations/HomeScreenGridPreferences.kt`

## Summary by CodeRabbit

* **New Features**
  * The home screen grid “rows” limit now adjusts automatically to the device’s screen size for a better fit across displays.
  * The rows slider maximum is now calculated dynamically and updated in both folded and unfolded preference views.
  * Any existing saved “rows” value is clamped to the current allowable maximum to prevent oversized layouts.

### DIFF
--- a/lawnchair/src/app/lawnchair/ui/preferences/destinations/HomeScreenGridPreferences.kt
+++ b/lawnchair/src/app/lawnchair/ui/preferences/destinations/HomeScreenGridPreferences.kt
@@ -1,5 +1,6 @@
 package app.lawnchair.ui.preferences.destinations
 
+import android.content.Context
 import android.content.res.Configuration
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -32,6 +33,7 @@ import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
+import androidx.window.layout.WindowMetricsCalculator
 import app.lawnchair.DeviceProfileOverrides
 import app.lawnchair.preferences.asPreferenceAdapter
 import app.lawnchair.preferences.getAdapter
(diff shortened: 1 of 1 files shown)


Broikos-Nikos:
Thanks for the review! Addressed both nitpicks in the latest push:

1. **Memoize the calculation** — done. Wrapped the `maxRows` selection in `remember(increaseMaxGridSize.state.value)` so it no longer recomputes on every recomposition / slider-drag frame.

2. **Window bounds instead of `displayMetrics`** — done, but using `computeMaximumWindowMetrics(context)` rather than `computeCurrentWindowMetrics()`. The home screen workspace is always rendered **fullscreen**, so its row capacity should reflect the whole display, not the (possibly shrunken) split-screen/multi-window bounds the settings screen happens to be in at the time. `getCurrentWindowMetrics()` would track that transient window and under-count on a device that's currently in split-screen. `computeMaximumWindowMetrics()` gives the true full-display size independent of multi-window state, which also fixes the underlying concern that an Activity's `resources.displayMetrics` can reflect the current multi-window bounds since Android R. `androidx.window` is already a dependency and `WindowMetricsCalculator` is already used in the widgetpicker module, so no new dependency was added.

MrSluffy (approved):

Broikos-Nikos:
Clamped the persisted `rows` value to the dynamic `maxRows` via a `LaunchedEffect(maxRows)`, so the slider never starts outside its own `3..maxRows` range (e.g. after disabling the experimental max-grid toggle on a smaller device) and can't silently truncate on the first nudge. Thanks — this was surfaced by CodeRabbit while reviewing a sibling branch.

SuperDragonXD:
Hello, and apologies for the delay in review. After looking at this, we've decided not to merge this PR as this change makes the slider behavior inconsistent across different devices. In other words, a simple toggle increasing the maximum count from 10 to 20 is more deterministic and better from a UX standpoint than trying to check the device size and changing the max value.

Feel free to check out other open issues if you're interested in tackling something else!

PR #6955, opened 2026-07-04
Add a separate icon pack option for the app drawer

Broikos-Nikos:
## What

Adds an optional **separate icon pack for the app drawer**, like Nova. A new **"Different app drawer icons"** switch (default **off**) lets the app drawer use its own icon pack while the home screen (and everything else) keeps the global one. When off, behaviour is unchanged.

## Why it's built this way

Lawnchair resolves icons **globally** — one `BitmapInfo` per component in a shared `IconCache` (backed by an external lib) with **no surface dimension** — and home, hotseat, taskbar, folders and search all read that one bitmap. The existing themed-icon split works only because both color and mono layers are baked into that single bitmap; two different *packs* can't share it.

So the global/home pack is left completely untouched and only the **drawer** diverges, via a render-time override:

- A memory-only `DrawerIconCache` bakes drawer-pack icons **lazily, per component, off the UI thread**, reusing the exact same recipe as the normal cache (`LawnchairActivityCachingLogic.loadIconForPack`) so badging, adaptive shaping, archived and themed layers all match. Components that can't be baked are remembered so they aren't retried, and newly installed apps pick up a drawer icon on first render.
- `BubbleTextView` (all-apps grid, predictions row, drawer folders, drawer app-search) and `PreviewItemManager` (closed drawer-folder previews) swap in the drawer bitmap when the toggle is on, via a new `ItemInfoWithIcon.newIcon(..., overrideBitmap)` overload.
- `LawnchairIconProvider` is parameterized by pack (`getIconForPack`) so the drawer reuses the same override → calendar → pack → themed resolution as the home path.
- The home icon-DB freshness key is intentionally untouched, so changing the drawer pack never rebuilds home icons.

## UI

A **"Different app drawer icons"** switch plus, when enabled, a third **"App drawer"** tab on the Icon Style screen for picking the drawer pack. The General → Icons subtitle reflects both packs when a separate drawer pack is chosen.

## Behaviour / scope

- **Default off → no change** for existing users; new prefs default to off/unset and never touch existing settings.
- **Zero overhead when off:** the cache is never instantiated and home/workspace icon binding does no extra work.
- **Drawer surfaces covered:** all-apps grid, predictions row, open + closed drawer folders, drawer app-search results, work profile / private-space app tabs. Private-space lock/header and non-app items are left alone.
- Dropping an app from the drawer to the home screen uses the home pack, as expected.
- Under memory pressure the drawer cache is dropped and lazily re-baked.

### Known limitations
- A dynamic calendar/clock icon **from the drawer pack** is baked once and won't roll over at midnight until the drawer pack/toggle changes (the global cache still refreshes normally).
- Themed drawer icons follow Lawnchair's existing rule that themed icons require the global "themed icons" setting; this is inherited behaviour, identical to using that pack globally.

## Files

Core: `icons/DrawerIconCache.kt` (new), `icons/LawnchairIconProvider.kt`, `LawnchairActivityCachingLogic.kt`, `preferences/PreferenceManager.kt`, `BubbleTextView.java`, `folder/PreviewItemManager.java`, `model/data/ItemInfoWithIcon.java`, `dagger/LauncherBaseAppComponent.java`. UI: `ui/preferences/destinations/IconPackPreferences.kt`, `GeneralPreferences.kt`, `res/values/strings.xml`.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting a separate icon pack for the app drawer.
  * Updated icon previews and folder items to reflect drawer-specific icons where enabled.
  * Improved icon pack labels and subtitles in settings to show both home and drawer selections.

* **Bug Fixes**
  * Drawer icons now refresh independently when their pack changes.
  * App drawer icons are cached and reloaded more reliably without affecting the home icon pack.

### DIFF
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -368,6 +368,10 @@
     <string name="themed_icons_home_and_drawer_label">Home screen and app drawer</string>
     <string name="lawnicons_not_installed_description">No supported icon packs</string>
 
+    <string name="separate_drawer_icon_pack_label">Different app drawer icons</string>
+    <string name="separate_drawer_icon_pack_description">Use a separate icon pack for the app drawer</string>
+    <string name="home_and_drawer_icon_pack_subtitle">%1$s (home), %2$s (drawer)</string>
+
     <!-- Fonts -->
     <string name="pref_fonts_add_fonts">Add fonts</string>
     <string name="pref_fonts_add_fonts_summary">OTF and TTF fonts are supported</string>
--- a/lawnchair/src/app/lawnchair/LawnchairActivityCachingLogic.kt
+++ b/lawnchair/src/app/lawnchair/LawnchairActivityCachingLogic.kt
@@ -7,7 +7,9 @@ import android.os.Build
 import android.os.Build.VERSION
 import android.os.UserHandle
 import android.util.Log
+import app.lawnchair.icons.LawnchairIconProvider
 import app.lawnchair.icons.getCustomAppNameForComponent
+import app.lawnchair.icons.iconpack.IconPack
 import app.lawnchair.preferences.PreferenceManager
 import com.android.launcher3.Flags.useNewIconForArchivedApps
 import com.android.launcher3.dagger.ApplicationContext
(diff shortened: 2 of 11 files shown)


Broikos-Nikos:
Thanks for the review! Addressed both drawer-related comments, and fixed a branch issue:

1. **`ComponentCallbacks2` never unregistered (Major)** — done. Stored the callback in a `memoryCallbacks` field, register it in `init`, and `unregisterComponentCallbacks(memoryCallbacks)` in `close()` before clearing.

2. **Drawer bitmap fetched for pending/promise icons (nitpick)** — done. `PreviewItemManager.setDrawable` now computes a `useDrawerPack` flag up front and only calls `DrawerIconCache.getBitmapInfoOrNull(...)` inside the branches that actually render an app icon, so active pending icons and app pairs no longer trigger a lookup/bake.

3. **`HomeScreenGridPreferences.kt` (rows clamp)** — that file belongs to an unrelated change (the dynamic max-rows feature) that was accidentally included when this branch was cut. I've rebased the branch so this PR now contains **only** the app-drawer changes; `HomeScreenGridPreferences.kt` is no longer part of the diff. That clamp suggestion is valid and is being handled on the other PR where the file actually changes.

Broikos-Nikos:
@coderabbitai review

Note: the branch was force-pushed after a rebase. The `ComponentCallbacks2`/`close()` unregister and the `PreviewItemManager` deferred-lookup comments are already addressed in the current head (`420e9183`), and `HomeScreenGridPreferences.kt` is no longer part of this PR.

SuperDragonXD (commented):
Hello! Just a quick note that we'll look into this PR, along with the other larger feature-related PRs, once Lawnchair 16 is released.

Also, please ensure that the checks pass here by running `./gradlew spotlessApply` to fix the lint errors.

Thanks for your patience!

PR #7031, opened 2026-07-18
fix: Restore default icon after resetting a custom override

pvolkov:
### Description

Fix resetting a per-app custom icon (from an icon pack) so the icon returns to the default/pack icon instead of staying stuck on the previous override. Applies to both personal and work profiles.

### Reasoning

After choosing a custom icon, reset deleted the override from the database, but the icon cache still treated the entry as fresh because `getStateForApp` did not include per-app overrides. Reloading via `forceReload()` could also skip `PackageUpdatedTask`, so the stale bitmap stayed cached. The fix includes override state in cache freshness, updates the in-memory map before refresh, and refreshes icons with `onAppIconChanged` while the model is still loaded.

### Testing

1. Long-press an app → customize → pick an icon from any icon pack (try one in **Personal** and one in **Work**).
2. Confirm the custom icon appears in the drawer/home.
3. Open customize again → **Reset to default**.
4. Confirm the icon immediately reverts to the normal/pack icon (not the custom one).
5. Optional: Settings → Home → **Reset custom icons** if multiple overrides exist; all should revert.

### Type of change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app icon refresh reliability after setting, resetting, or deleting custom icons.
  * Reduced chances of stale icon cache results when icon overrides change.
  * Ensured clearing all icon overrides immediately updates the current icon state.
  * Added more accurate per-app icon override state tracking so cached icons refresh when customizations are updated.

### DIFF
--- a/lawnchair/src/app/lawnchair/data/iconoverride/IconOverrideRepository.kt
+++ b/lawnchair/src/app/lawnchair/data/iconoverride/IconOverrideRepository.kt
@@ -1,6 +1,7 @@
 package app.lawnchair.data.iconoverride
 
 import android.content.Context
+import android.os.UserHandle
 import app.lawnchair.data.AppDatabase
 import app.lawnchair.icons.picker.IconPickerItem
 import com.android.launcher3.LauncherAppState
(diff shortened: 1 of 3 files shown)


validcube (approved):
LGTM

PR #7075, opened 2026-07-28
fix: prevent clock widgets crashing on Android 14

### LINKED ISSUE #7062: Adding widgets crashes LC-16
### Describe the bug

I've been trying out lawnchair 16 and have stumbled upon some bug that makes the launcher crash when trying to add widgets to the launcher.
This bug was experienced on a Samsung Galaxy A52 4g running the latest available official Samsung software. 

### Steps to reproduce

1. Hold down the homescreen
2. Press "widgets"
3. Drag any widget from (from my experience) any app onto the homescreen
4. Launcher crashes

### Expected behavior

The widgets gets added to the homescreen. 

### Screenshots

Not applicable 

### Device information

- Device: Samsung Galaxy A52 4G (SM-A525F/DS) 
- Android version: 14

### App version

16.Dev (#4691)

### Additional context

Image

mvanhorn:
### Description

Adding a widget whose `RemoteViews` contains a `TextClock` or `AnalogClock` crashes Lawnchair 16 during asynchronous layout inflation on Android 14. The crash log in the issue shows `QuickstepLauncher.onCreateView` resolving `android.widget.TextClock.ClockEventDelegate`, which is absent from the reporter's API 34 Samsung framework.

This changes `QuickstepLauncher.onCreateView` to install `AsyncClockEventDelegate` only on Android 15 / API 35 and newer, letting the normal framework inflater path create clock views on Android 14 and older. `AsyncClockEventDelegate`'s documentation and API annotations move to the same API 35 boundary, so callers and static analysis cannot treat the hidden framework type as available on API 34.

Fixes #7062

### Reasoning

`ClockEventDelegate` is a hidden framework type that only appears in the API 35 SDK sources, but the current `16-dev` code gates on `Utilities.ATLEAST_U` (API 34). On an API 34 device the class does not exist, so the async-inflation path throws while binding the widget.

The repository already defines `Utilities.ATLEAST_V`, so the fix is to move the gate to the constant that matches where the framework type actually exists, rather than adding a reflective probe or a try/catch around inflation. Keeping the annotation and the runtime gate at the same API level means lint can enforce the boundary instead of relying on the call site staying correct.

### Testing

- Added an instrumentation fixture, `tests/res/layout/test_layout_appwidget_clock.xml`, containing a clock view, and extended `BindWidgetTest` to bind a widget using it. That exercises the same activity layout-factory wiring used when a widget is placed on the workspace, which is the path that crashed.
- The new coverage sits in the existing `BindWidgetTest` suite, so it runs with the project's normal instrumentation test task; no new harness was introduced.
- Not reproduced on a physical API 34 Samsung device. The fix targets the API boundary named in the reporter's stack trace, so confirmation from the reporter on that hardware would be valuable.

### Type of change

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved clock widget handling on Android 15 and later.
  - Maintained compatibility with older Android versions through an appropriate fallback.
  - Improved asynchronous loading of clock widgets containing digital and analog clock elements.

- **Tests**
  - Added coverage to verify clock widgets inflate and appear correctly when loaded asynchronously.


validcube (commented):
LGTM. Can you remove the changes in tests folder? We don't use tests at the moment.

mvanhorn:
a280a7e takes the tests/ changes back out, so the diff is just the widget fix now. That removed the clock test along with its three imports and the test-only layout it referenced.

validcube (approved):

PR #7097, opened 2026-08-04
fix: refresh configured icon gestures

MuonChaser:
## Description

Per-app icon gestures were cached when an existing workspace icon first checked them. Configuring a gesture later did not affect that icon until it was removed and added again, which created a new listener.

Resolve the configured state from `PreferenceManager2`'s in-memory cache on each check so an existing icon observes updated gesture settings immediately.

## Testing

- `git diff --check`
- `./gradlew projects`
- Not run: Android Kotlin compilation requires a local Android SDK, which is not available in this environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Gesture settings changes now take effect immediately without requiring icon views to be recreated.
  * Improved detection of configured gestures after per-app settings are updated.
  * Gesture configuration is refreshed more efficiently, improving responsiveness during scrolling.

### DIFF
--- a/lawnchair/src/app/lawnchair/gestures/IconGestureListener.kt
+++ b/lawnchair/src/app/lawnchair/gestures/IconGestureListener.kt
@@ -32,10 +32,6 @@ class IconGestureListener(
      */
     private enum class ScrollLock { NONE, HORIZONTAL, VERTICAL }
 
-    private val configuredGestures by lazy {
-        GestureType.entries.filter { resolveGesture(it) != null }.toSet()
-    }
-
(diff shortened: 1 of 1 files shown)


MuonChaser:
It's a small bug to be fixed, while there's still other issues when swiping up the icon is comfilct with entering app drawer. I'm quite keen on the icon gestures, but their behaviours are uneasy to predict now.

MuonChaser:
@coderabbitai resume
@coderabbitai review

validcube (approved):

PR #7115, opened 2026-08-06
feat: Enable private space apps on the home screen

alfs:
### Description

This patch adds user opt-in pinning of apps in private space. When the private space is unlocked, apps and folders are shown. When the private space is locked, the apps are hidden and vice versa.

### Reasoning

Private space apps can normally not be pinned to the home screen due to being flagged FLAG_NOT_PINNABLE, and any app that (due to bugs) did end up in the homescreen were deleted whenever the private profile was not visible (looked like a deleted app so remove from screen) or could not be moved/dragged. Even if private space apps are meant to be kept separate and off the home-screen, it makes sense from a usability perspective to allow the user to pin such apps as well, or put them in folders. At the same time, the private properties should be preserved when the private space is locked.

Why they were being deleted

Access to the private profile comes from ACCESS_HIDDEN_PROFILES, an appop permission held only while this app has the home role. Losing that role - or simply locking the space - makes the profile disappear from UserCache and makes LauncherApps report every package in it as absent.  That is indistinguishable from the app having been uninstalled, so the loader deleted their database rows
- each pinned item is one row of the favorites table, holding its profile, position and intent. Switching launchers and back, or rebooting with the space locked, was enough to lose every pinned private app with no way to recover the positions.

PrivateProfileTracker records the profile's serial and user id while it is visible, so those database rows stay resolvable when it is not. The record is dropped only when the profile is genuinely gone, which UserManager can still confirm by serial even for a profile LauncherApps will not show.

Features

- Opt-in "Allow on home screen" preference (default off) that releases FLAG_NOT_PINNABLE for private space apps.
- Pinned private apps are hidden while the space is locked, keeping their database rows and reappearing on unlock.
- Tapping a pinned private app while locked asks the system to unlock the space rather than failing silently, prompting for the home role if we do not hold it.
- Optional "Hide app badge" preference to drop the private profile badge from these icons.
- Backups made while the space is unlocked restore pinned private apps in full.

Privacy considerations

Hiding an app is only as good as its least careful presentation, so every surface that reaches the model independently is filtered: the workspace, the taskbar, folder previews, app pairs, drawer folders and search. Both search algorithms enumerated the locked space by name, since all apps keeps private apps loaded across a lock and only the drawer adapter filtered them.

A folder or app pair is hidden whole when any of its contents is hidden, rather than having its contents filtered.

Filtering was tried first and produced a string of states the launcher never normally reaches. Launcher3 collapses a folder into a plain icon as soon as it holds one item, so a folder left showing a single app is an anomaly an observer can read as "something is hidden here" - as is an empty slot inside an open folder. Fixing each symptom took a separate patch in a separate file, and they interacted: moving one app out of a mixed folder made an unrelated normal app disappear, and whether a folder showed at all depended on how many private apps it held relative to normal ones. Hiding the container whole replaced all of it with one rule that also covers app pairs, and left three files untouched.

The trade is that a folder holding one private app and ten normal ones disappears entirely while the space is locked, where filtering would have shown the ten. That is worse for that case but predictable, and avoidable by not mixing private and normal apps in the same folder. The apps in a hidden folder remain reachable from the drawer.

Hidden icons are not bound at all rather than made invisible. The problem is that an invisible child still holds its cell and silently refuses drops, which would reveal how many private apps are pinned and where. Cells are left free instead, and an icon whose cell was taken while the space was locked is relocated on unlock - never deleted - with a toast naming it only when it is actually visible.

Backup export scrubs private rows from launcher.db when the space is locked, since the favorites table stores component names in plain text and the file leaves the device.

### Testing
Setup

  Install app, set it as default home app, then enable Home screen → Private space → Allow on home screen.

  The default-home step isn't optional since ACCESS_HIDDEN_PROFILES is granted only to the role holder, so without it every private app looks absent

  adb logcat -s PrivateProfileTracker PrivateSpacePlacement WorkspaceItemProcessor LoaderCursor PrivateSpaceUtils

  0. Locking survival

  Lock the space, force an app reload. Watch for PrivateProfileTracker saying the recorded profile no longer exists. That must
  appear only when you actually delete the private space. If it fires while merely locked, the patch deletes the database rows it exists to protect — which shouldnt happen.

  1. Reboot survival

  Pin two or three private apps, lock, reboot, unlock. All should return in place. Failure looks like
  WorkspaceItemProcessor logging "Invalid package removed". Drag and put in folder. Reboot and check again.

  2. Launcher switch
  
  Switch default home away and back. This causes the appop to be removed and reinstated. Apps should return in place as in step 1.

  3. Relocation 
  
  Note a private icon's cell. Lock (it disappears, cell becomes droppable). Drop a normal app there. Unlock.

  Expect: normal app keeps the cell, private app moved, toast informs. If the normal app vanishes instead, the deferral 
  ordering is backwards. Turn Allow overlap off first or this test does nothing.

  4. Grid size change

  Change grid size, locked and unlocked. Private icons must survive both — this was a separate bug from the reboot one.

  5. Privacy surfaces, with the space locked

  Each of these surfaces reaches the model independently, so check them separately: home screen, dock/taskbar, drawer search by name (must return nothing), and drawer folders.

  Folders — any folder containing a private app disappears whole while locked and
  returns intact on unlock. No single-app folders, no gaps inside open folders. If you see either, the rule isn't being
  applied where I think it is.

  6. Backup round trip

  Export locked → unzip and confirm launcher.db has no private rows. Export unlocked → restore → private pins return.

  7. Auto-add and the stuck-icon report

  Preference off: installing an app into the private space must not auto-add it, and existing private icons stay
  undraggable. Preference on: they become draggable on the next load.

### Type of change

## Summary by CodeRabbit

* **New Features**
  * Added Android 15+ Private Space settings for app pinning and hiding Private Space badges.
  * Private Space apps can remain available on the home screen when configured, with automatic relocation notices when needed.
  * Locked Private Space apps can prompt profile unlocking when selected.
* **Bug Fixes**
  * Locked Private Space apps are hidden from app drawers, taskbars, and search results.
  * Private Space apps and data are better preserved during backup, restore, and launcher grid migration.

### DIFF
--- a/AndroidManifest-common.xml
+++ b/AndroidManifest-common.xml
@@ -43,7 +43,14 @@
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.EXPAND_STATUS_BAR" />
     <uses-permission android:name="android.permission.ACCESS_RESTRICTED_SETTINGS" />
-    
+
+    <!--
+    Required to see the private profile at all. This is an appop permission granted only while this
+    app holds the home role, so it comes and goes as the default launcher changes. Declared here
+    rather than only in the quickstep manifest so that every flavour can observe the profile.
+    -->
+    <uses-permission android:name="android.permission.ACCESS_HIDDEN_PROFILES" />
+
     <!-- for rotating surface by arbitrary degree -->
     <uses-permission android:name="android.permission.VIBRATE"/>
     <uses-permission android:name="android.permission.ROTATE_SURFACE_FLINGER" />
--- a/src/com/android/launcher3/model/PrivateSpacePlacement.kt
+++ b/src/com/android/launcher3/model/PrivateSpacePlacement.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.android.launcher3.model
+
+import android.content.ContentValues
+import android.provider.BaseColumns
+import android.util.LongSparseArray
+import android.widget.Toast
+import app.lawnchair.util.PrivateSpaceUtils
+import com.android.launcher3.LauncherAppState
+import com.android.launcher3.LauncherSettings.Favorites
+import com.android.launcher3.R
+import com.android.launcher3.logging.FileLog
+import com.android.launcher3.model.data.ItemInfo
+import com.android.launcher3.util.Executors.MAIN_EXECUTOR
+import com.android.launcher3.util.IntArray as LauncherIntArray
+
+/**
+ * Workspace placement for private space items, kept out of [LoaderTask] so that the loader carries
+ * only the call sites.
+ *
+ * While the private space is locked its pinned icons are not shown, so their cells look empty and
+ * the user may drop something onto one. Two rows then name the same cell, and the loader's ordinary
+ * overlap handling would delete whichever it reached second - by cursor order, that is usually the
+ * item the user just placed. So private items claim no cell during the cursor pass; they are placed
+ * here afterwards, once every visible item has settled, and relocated rather than deleted when their
+ * cell has gone.
+ */
+object PrivateSpacePlacement {
+
+    private const val TAG = "PrivateSpacePlacement"
+
+    /**
+     * Gives a final position to items that were read without claiming a cell.
+     *
+     * An item keeps its stored cell whenever that cell is still free; otherwise it moves to the
+     * first free one and the move is written back so it survives the next load. Deleting instead
+     * would be unrecoverable, as the app cannot be re-pinned until the space is unlocked.
+     */
+    @JvmStatic
+    fun placeDeferredItems(app: LauncherAppState, dataModel: BgDataModel, c: LoaderCursor) {
+        val deferred = c.deferredPlacements
+        if (deferred.isEmpty()) return
+
+        val firstPagePinned = dataModel.isFirstPagePinnedItemEnabled
+        val screens = dataModel.collectWorkspaceScreens()
+        val addedScreens = LauncherIntArray()
+        val spaceFinder = WorkspaceItemSpaceFinder()
+        val relocated = mutableListOf<CharSequence>()
+
+        for (info in deferred) {
+            // The whole body is guarded: this runs per item during the workspace load, and letting
+            // anything escape would abort the load rather than lose a single icon.
+            try {
+                if (c.checkItemPlacement(info, firstPagePinned)) continue
+                val coords = spaceFinder.findSpaceForItem(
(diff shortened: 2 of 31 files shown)


alfs:
@coderabbitai review

src/com/android/launcher3/provider/RestoreDbTask.java
```diff
@@ -253,6 +254,15 @@ protected int sanitizeDB(Context context, ModelDbController controller, SQLiteDa
...
                 FileLog.d(TAG, "sanitizeDB: managed profile id=" + oldManagedProfileId
                         + " should be mapped to new id=" + newManagedProfileId);
+            } else if (PrivateProfileTracker.isRestorablePrivateProfileSerial(
+                    context, oldManagedProfileId)) {
+                // Private profiles are never part of backup & restore, so BackupManager cannot map
+                // them and every private space item would be deleted below. When the serial still
+                // belongs to a private profile on this device - always the case for a local
+                // Lawnchair backup - the rows are still valid, so map the id to itself.
+                profileMapping.put(oldManagedProfileId, oldManagedProfileId);
+                FileLog.d(TAG, "sanitizeDB: keeping items of private profile id="
```
MrSluffy, line 264: don't use FileLog.d

src/com/android/launcher3/model/WorkspaceItemProcessor.kt
```diff
@@ -194,6 +201,16 @@ class WorkspaceItemProcessor(
                 // Points to a valid app (superset of cn != null) but the apk
                 // is not available.
                 when {
+                    isInaccessiblePrivateItem -> {
+                        FileLog.d(
```
MrSluffy, line 205: this one too

src/com/android/launcher3/util/PrivateProfileTracker.kt
```diff
@@ -0,0 +1,294 @@
...
+            // Also a hidden API. UserHandle.hashCode returns the same user id, so it is a safe
+            // fallback rather than a reason to give up recording the profile.
+            user.hashCode()
+        }
+        val prefs = LauncherPrefs.get(context)
+        if (
+            prefs.get(KNOWN_PRIVATE_PROFILE_SERIAL) != serial ||
+                prefs.get(KNOWN_PRIVATE_PROFILE_USER_ID) != userId
+        ) {
+            FileLog.d(TAG, "Recording private profile serial=$serial userId=$userId")
```
MrSluffy, line 101: this one too

src/com/android/launcher3/util/PrivateProfileTracker.kt
```diff
@@ -0,0 +1,294 @@
...
+    fun onProfileRemoved(context: Context, user: UserHandle) {
+        val prefs = LauncherPrefs.get(context)
+        val known = prefs.get(KNOWN_PRIVATE_PROFILE_SERIAL)
+        if (known == INVALID_SERIAL) return
+        // UserCache refreshes asynchronously, so the removed user is usually still resolvable here.
+        // Its fallback path also reports a user carrying the recorded serial as private, which
+        // covers the case where the cache has already dropped it.
+        val info = UserCache.getInstance(context).getUserInfo(user)
+        if (info.isPrivate || info.userSerial == known) {
+            FileLog.d(TAG, "Private profile removed, forgetting serial=$known")
```
MrSluffy, line 122: uhhh

src/com/android/launcher3/util/PrivateProfileTracker.kt
MrSluffy, line 1: move this to lawnchair module folder

AndroidManifest-common.xml
MrSluffy, line 1: can be add in lawnchair module manifest

src/com/android/launcher3/model/PrivateSpacePlacement.kt
MrSluffy, line 1: for all new file should be in lawnchair moduel

MrSluffy (changes requested):

PR #6808, opened 2026-05-28
feat: Make application sorting in Deck mode optional

lebao3105:
### Description

Fixes #6737 and #6807 (not all of the requested features right now)

### Testing

Visual test

(backup all settings before testing)

### Type of change

:x: **Bug fix** (A non-breaking change that fixes an issue)
:white_check_mark: **New feature** (A non-breaking change that adds functionality)
:x: **Breaking change** (A fix or feature that would cause existing functionality to not work as expected)
:x: **Refactor** (A code change that neither fixes a bug nor adds a feature)
:x: **Performance** (A code change that improves performance)
:x: **Style** (Code style changes)
:x: **Docs** (Changes to documentation)
:x: **Chore** (Changes to the build process or other tooling)

## Summary by CodeRabbit

* **New Features**
  * Added a “Categorize apps” option in Deck mode to organize apps into category folders.
  * Newly installed apps can be automatically placed into matching categories when categorization is enabled.
  * Deck mode now shows progress updates for restoring layouts, adding apps, and creating folders.
  * Added a preference to enable or disable app categorization.

* **Bug Fixes**
  * Improved Deck mode restoration and home screen refresh consistency.
  * Updated Deck mode disabling to restore swipe-up access to the app drawer with more reliable reload handling.

PR #7125, opened 2026-08-08
fix(deck): promote to stable the lawndeck

MrSluffy:
### Description

Promotes the Deck (drawerless) layout from Experimental Features to a stable Home Screen preference, and fixes the bugs that were blocking promotion: a deadlock that could leave the enable dialog stuck forever, an arbitrary 800 ms wait replaced with deterministic completion, empty/duplicated folders, missing folder item ranks, backups breaking when the grid size changed, and newly installed apps (Play Store or sideloaded) not appearing in the deck after re-enabling it.

### Reasoning

- **Deterministic population.** Deck population previously read apps from the launcher activity's UI (`mAppsView.appsStore`) from an IO thread and waited on `ItemInstallQueue` with a `postDelayed(800)` hack; early returns never completed the awaited deferred, hanging the blocking dialog. The logic now lives in a single model task (`AddAllAppsToWorkspaceTask`) that reads `AllAppsList` on the model thread, places folders *and* single apps itself, and always invokes its completion callback (`finally`). Enqueueing is guarded by `LauncherModel.loadAsync`, because `enqueueModelUpdateTask` silently drops tasks while the model is unloaded (which it is right after a DB restore).
- **Correct folder contents.** Items are filtered through `shortcutExists` *before* folders are written to the DB (no more persisted empty folders), `item.rank` is now set (folder order is rank-driven), and `addNewlyInstalledApp` had an off-by-one position and no duplicate check.
- **Grid-safe backups.** Backups now snapshot every grid DB (both AOSP `launcher_X_by_Y.db` and Lawnchair `launcher_R_C_H.db` naming) plus a sidecar recording the `DeviceGridState`. If the grid changed while a snapshot was inactive, restore writes the recorded state to the `migration_src_*` prefs so launcher migrates the restored layout into the current grid through its native migration path. Stale SQLite `-journal`/`-wal`/`-shm` files are deleted before restoring (a leftover journal can roll back a restored DB), and the regular-layout snapshot is refreshed on every enable so disabling restores the layout you actually had. Old single-file backups still restore.
- **Reconcile on re-enable.** The populate task now also runs after restoring a deck snapshot; since existing items are skipped, it adds exactly the apps installed while the deck was off, appending them to existing category folders instead of creating duplicates. Apps uninstalled while the deck was off are pruned by the loader on reload.
- **Promotion.** The `showDeckLayout` gate toggle, preference, and its strings (base + 57 locales) are removed; the Deck layout option now always appears in Home Screen preferences, and the "(Beta)" label is dropped. Progress messages are now localized string resources instead of hardcoded English.

### Testing

1. Home Screen preferences → toggle **Deck** on: apps are categorized into folders (Google Apps, System Apps, Flowerpot categories); single-app categories appear as plain icons; the progress dialog always dismisses.
2. Install an app (Play Store *and* sideload) while deck is on: it lands in the matching category folder, appended last, or on the workspace if no folder matches. No duplicates.
3. Toggle deck off: the previous regular layout is restored. Rearrange some icons, toggle on/off again: the *latest* regular layout comes back, not the original one.
4. While deck is off, install one or more apps, then re-enable deck: the new apps appear in the restored deck (existing folders reused, no duplicate folders). Uninstalled apps are gone after reload.
5. Enable deck → disable → change the home screen grid size → re-enable → disable: the layout follows the grid change via launcher's native grid migration; no empty or stale workspace.
6. Verify Experimental Features no longer shows the "Deck layout" toggle and the option in Home Screen preferences is labeled "Deck" (no "Beta").

### Type of change

## Summary by CodeRabbit

- **New Features**
  - Added automatic app categorization when setting up the Lawndeck layout.
  - Apps are added directly or grouped into folders, while existing shortcuts are preserved.
  - Added progress messaging for restoring the layout, organizing apps, and adding apps to the home screen.
  - Improved layout restoration with backup support and cleanup of outdated data.

- **Bug Fixes**
  - Prevented duplicate app shortcuts during layout setup.
  - Simplified home screen layout settings and removed the obsolete experimental layout toggle.
  - Updated localized strings across supported languages.


lebao3105:
You may also want to implement #6808, as I'm not capable of fixing existing problems (if any)

PR #6233, opened 2025-12-29
feat(widget-stacks): initial implementation of widget stacking

MrSluffy:
### Description

This feature implements widget stacking functionality similar to Samsung One UI 5.0, allowing users to stack multiple widgets in the same space on the home screen.

### Fixes
- Closes: #3223 

### TODO
- [x] Fix live widget when app restart
- [x] Fix preview on template
- [x] Resize stack widget similar with the single widget
- [x] Reorder widgets within stacks via reorder pref
- [x] UI Widget Picker

### Preview
Screenshot_20260411-182623_Lawnchair
Screenshot_20260411-182629_Lawnchair

### Type of change

:x: **Bug fix** (A non-breaking change that fixes an issue)
:white_check_mark: **New feature** (A non-breaking change that adds functionality)
:x: **Breaking change** (A fix or feature that would cause existing functionality to not work as expected)
:x: **Refactor** (A code change that neither fixes a bug nor adds a feature)
:x: **Performance** (A code change that improves performance)
:x: **Style** (Code style changes)
:x: **Docs** (Changes to documentation)
:x: **Chore** (Changes to the build process or other tooling)

## Summary by CodeRabbit

* **New Features**
  * Widget stacks: combine multiple widgets into one home-screen cell with add/remove/reorder support.
  * Full-screen stack editor with widget picker, drag-to-reorder pages, and page indicators.
  * Auto-rotate stacked widgets and per-stack rotation toggle.
  * Drag-to-stack creation (drop-to-create / add-to-existing) and long-press menu to create/edit stacks.
  * Stack-aware widget resizing and persistence so stacks survive restarts.

Issue #3223, opened 2023-01-02
[FEATURE] Stacked Widgets


RydahDoesTech:
### Describe the feature

Stacked Widgets allow for multiple widgets to be placed within one container. They can be accessed by scrolling within the container. This helps users use less home screen pages yet all the widgets they need.

### Why would it be useful to add?

This would allow users to have more widgets on less home pages. Users would be able to create stacks on their home screen and have multiple widgets within them that they can scroll between.

### Did the feature exist in Lawnchair v2? (Play Store version)

Yes

### Example(s)

https://media.discordapp.net/attachments/803506450805817385/1059567144804823050/Screen_Recording_20230102_122121_One_UI_Home.mp4

https://media.discordapp.net/attachments/803506450805817385/1059570943942467606/Screen_Recording_20230102_123627_One_UI_Home.mp4

### Additional context

This feature is present in Samsung's OneUI Home and Smart Launcher 6.

RWCoding58:
Might be a little difficult to implement, but I'm definitely all for it being added in the future.
**Recommendation:** This is a clear, desirable feature request that fits the project vision.

RWCoding58:
Feature is basically to add widget stacking support, similar to like in [One UI](https://www.samsung.com/us/support/answer/ANS10000945) and [iOS](https://support.apple.com/en-us/118610#:~:text=How%20to%20use%20widget%20stacks). Will probably need to consider conflicting widget sizes and possibly add constraints (like how One UI only allows widgets of the same size to be stacked).

RWCoding58:
Uh really sorry, but it seems like widget stacking probably will not be added any time soon due to complexities (see [here](https://github.com/LawnchairLauncher/lawnchair/blob/15-dev/ROADMAP.md#long-term-or-blocked)).
I'll leave the issue open but I will not add it to tracked for now.

PR #7156, opened 2026-08-13
feat: widget stacking with axis-locked paging and home wrap

SotirisPs:
### Description

Adds widget stacking on the home screen (recovering #6233), with per-stack horizontal/vertical paging, in-direction wrap, a hold-to-move duration so the edit sheet is not covered by a drag ghost and it doesn't get effected by other gestures, and home infinite scrolling that wraps the same way instead of jumping all the way around.

### Reasoning

Official Lawnchair still lacks live widget stacks. This continues the unmerged stacking work and makes wrap feel like a normal page swipe because there isn't an option between a smooth wrap around and jump back to the beginning.

### Testing

- [ ] Add a widget stack, pick widgets, swipe only on the selected axis
- [ ] Wrap last → first and first → last; animation follows the finger and doesn't stater or warps its animation
- [ ] Long-press a widget/stack: menu opens; keep holding to enter move/resize (default 2.5 sec)
- [ ] Vertical stack: only up/down pages; left/right does not change the widget; up still opens the app drawer
- [ ] Wrap from last → first (and back) stays in the same swipe direction (loops around instead of jumping to the first widget)
- [ ] Last widget stays visible after landing on it
- [ ] Long-press for settings panel → Edit stack does not show a widget ghost on press

### Type of change


MrSluffy:
this is amazing Thanks for recovering this, will check and test this! please change the base to 16-dev!

HuangChengHsien:
Hi @SotirisPs,

I've been building a widget-stacking implementation on a private fork (patches on top of pinned Lawnchair, separate from this branch) and just reached a correctness-hardening baseline (v1.0) after several rounds of testing, so I wanted to share notes here since this PR (and #6233 before it) is the closest active work on the same feature.

A few things that turned out to matter, in case they're useful:

- Keeping both the stack root and every child as real Favorites rows (each with its own real `appWidgetId`), plus a small side-table for stack membership/order/active child, made cold start, backup/restore, and provider-uninstall recovery much easier to reason about than I expected.
- TalkBack accessibility needed explicit handling: suppressing the resize action for anything inside a stack, and making sure accessibility focus lands on the stack as one node rather than on nested children.
- Grid size / orientation migration needed to explicitly exclude stack children from occupancy checks and row-shift SQL, or it would read/rewrite their (intentionally stale) positions.
- Deleting a stack has to close out the launcher's own delete-undo bookkeeping afterward, or later unrelated deletes can silently stop taking effect.
- Provider-uninstall recovery (3→2 children keeps the stack, last child reverts the root to a standalone widget, root itself removed dissolves the whole stack) needed explicit reconciliation both on a full reload and while the launcher is already running, not just at bind time.

I noticed #6233 was closed without merging and this PR hasn't had activity in a while — is it still active, or has priority shifted elsewhere? Happy to compare notes on any of the above if useful, since #3223 clearly has a lot of demand behind it.

PR #7193, opened 2026-08-21
Add driving mode UI, per-shortcut icon customization, and widget/backup fixes

adegn73:
## Summary

This bundles a few related features and fixes developed together:

**Driving mode** - as Google has removed the Car Dock feature from Android, it is now necessarry to create a launcher, that can switch to a different - simpler UI if needed. The Driving Mode is inspired by Google Car Home:
- User-configurable rows/columns/pages, background color/image/opacity, matching the app drawer's background picker
- Long-press any tile to assign an app or a built-in quick action (Exit, Settings, Navigation, Music, Phone, Contacts, and a new live Speedometer with an mph/km-h setting)
- Fixes: driving mode not appearing until returning to the launcher manually, nav bar not picking up the driving-mode tint, Home button opening search instead of resetting the grid, and touch/scrolling breaking permanently after a rotation
- New quick settings tile to toggle driving mode, always reflecting its live state

**Per-shortcut icon customization** - pick a custom photo icon for individual shortcuts/apps, and optionally hide the work/clone-profile badge per shortcut

**App drawer** - paged grid style, background image/color support (shared with driving mode), font fix, background pans with paged-drawer scroll

**Widget corner-radius enforcement** - the "Rounded corners" setting was silently unwired (toggling it did nothing); renamed to "Use system radius" and wired up for real, plus a new "Custom radius" slider (1-25dp) for devices where the OS-reported system radius clips old/legacy widgets too aggressively

**Backup & restore** - background images and custom icon photos are now included in backups (previously silently lost on restore, since only the DataStore/Room DB files were captured, not the loose image files those features write separately)

## Test plan
Extensively tested on a physical device (Motorola Edge 70, Android 16) throughout development: driving mode paging/rotation/background/quick-actions/speedometer, the quick settings tile toggling correctly from multiple trigger paths, the widget radius fix against a real widget that was being clipped by the OS's system radius, and a full backup/restore round-trip verified by inspecting the zip contents directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added Driving Mode with Bluetooth activation, Quick Settings access, customizable app/action tiles, speedometer, backgrounds, layouts, and speed units.
  - Added a paged app drawer with configurable rows, infinite scrolling, and optional background images.
  - Added custom app and shortcut icons, optional shortcut badge hiding, and customizable widget corner radius.
  - Added an option to hide Lawnchair activities from Recents.
- **Bug Fixes**
  - Improved custom icon loading, replacement, and backup preservation.
  - Improved app drawer backgrounds and layout behavior across device configurations.
  - Improved font selection and icon refresh behavior.


SuperDragonXD (changes requested):
Hello, and thanks for your interest in contributing to Lawnchair!

Before anything else, please [review Lawnchair's contributing guidelines](https://docs.lawnchair.app/developers/introduction/get-started#creating-your-first-contribution) first, as the scope of this PR is quite large. Alongside that, driving mode is not something we plan on implementing, as it is quite out of scope with a *home screen app*.

We will properly look into this once Lawnchair 16 is released.

adegn73:
Hi @SuperDragonXD, thanks for taking a look and for the honest feedback — that's fair, and I appreciate you flagging the contributing guidelines. I'll split this into smaller, focused PRs going forward rather than bundling everything together.

On driving mode specifically: I completely understand why it looks out of scope for a home screen app at first glance, and I don't want to push back on that call — just wanted to explain the reasoning in case it's useful context for when you revisit this after 16 ships.

The short version is that Google no longer ships anything like this at the OS level. Android Auto for phone screens (the built-in fullscreen driving UI that used to activate automatically over Bluetooth) was killed off back in 2021–2022, and its thinner replacement, Assistant Driving Mode, has since been stripped down to almost nothing and is being retired entirely as Assistant gets replaced by Gemini. There's no first-party "press connect to car, get a driving UI" experience left on stock Android anymore.

The reason this ends up needing to live in a launcher rather than a standalone app is a platform restriction, not a design preference: reliably bringing a UI to the foreground automatically from a background trigger (a Bluetooth connection event, with no user tap) is exactly what Android's background-activity-launch restrictions are built to block, and the Home/launcher app is one of the few things exempted from that. A regular app just can't do this reliably anymore — the old AOSP CarHome trick of registering for CATEGORY_CAR_DOCK and having Home resolve to it during car mode isn't consistently honored on modern devices either, so a launcher-level implementation is really the only path left that works reliably across devices.

Given that, I think there's a real opportunity here: since no one — not Google, not any major launcher I'm aware of — currently offers this, it could be a genuine point of differentiation for Lawnchair rather than scope creep. Happy to resubmit it as its own standalone, well-scoped PR whenever the timing works better on your end, and take any direction on how you'd want it designed/gated (e.g. opt-in setting, separate module, etc.) to fit cleanly.

Thanks again for the review, and no worries either way — appreciate the project and the time you put into looking at this.

PR #7255, opened 2026-09-07
fix: Icon pack and drawer settings that could not be reached or did not apply

### LINKED ISSUE #7254: Icon pack and app drawer settings that are unreachable or do not apply
### Describe the bug

A handful of icon-pack and app-drawer settings on `16-dev` are either unreachable from the UI or do not apply to the surface they claim to. Some are regressions from the icon pack settings rework in ba62bdf380; the rest are settings that are read by code but can no longer be set.

**1. "Tint with accent color" is gone from the icon pack tab**

`tintIconPackBackgrounds` is still read by `LawnchairIconProvider`, but the switch that set it was dropped from the icon pack tab, so the preference can no longer be changed.

**2. The themed icon pack tab lists the wrong packs**

`IconPackGrid` stopped honouring `isThemedIconPack`, so the themed icon pack tab lists the regular icon packs. Those packs provide no monochrome icons, so picking one does nothing.

**3. The drawer's "suggested apps" switch has no UI**

`PredictionRowView` reads a preference that no screen can set any more — it was lost when predictions were reworked. The drawer's suggestion row cannot be turned off.

**4. Half of the icon style settings are invisible depending on the open tab**

The icon style page keeps each setting inside the tab whose pack it belongs to, so switching tabs hides settings that are not tab-specific.

**5. An app dragged out of the drawer keeps the plain system icon**

Dragging an app from the drawer to the home screen carries the icon the drawer drew, so the home screen shows the unthemed/unpacked icon until the next model reload, even though the icon pack does apply on the home screen.

**6. The drawer's background opacity setting does nothing on a phone in portrait**

The opacity is only applied on the path where the all-apps sheet is drawn without blur, so in portrait on a phone the slider has no visible effect.

**7. The first tap on the drawer search bar does not open the keyboard**

Taking focus runs the zero-state search, and rebuilding the list under the field drops the keyboard that the framework had already asked for. A second tap finds the field focused and keeps the keyboard, so it only ever opens on the second tap.

### Steps to reproduce

1. Settings → Home screen → Icon style → **Icon pack** tab → the "Tint with accent color" switch is not there (1).
2. Settings → Home screen → Icon style → **Themed icons** tab → the list shows normal icon packs, not `app.lawnchair.icons.THEMED_ICON` ones; select one and nothing is themed (2).
3. Switch between the icon style tabs and watch settings that are not about the selected pack appear and disappear (4).
4. Settings → App drawer → Suggestions → there is no switch for the drawer's suggested apps row (3).
5. Open the app drawer, long-press an app and drag it onto the home screen → the icon on the home screen is the plain system icon while the drawer still shows the icon-pack one. It corrects itself after a launcher restart (5).
6. On a phone in portrait, Settings → App drawer → change the background opacity → nothing changes (6).
7. Open the app drawer and tap the search bar once → the field takes focus but the keyboard does not come up; tap again → it does (7).

### Expected behavior

1. The tint switch is on the icon pack tab, as it was before ba62bdf380.
2. The themed icon pack tab lists only packs that respond to `app.lawnchair.icons.THEMED_ICON`.
3. The drawer's suggested apps row has a switch.
4. Icon style settings are visible regardless of which pack tab is open.
5. An app dropped on the home screen gets the home screen's icon straight away.
6. Background opacity applies to the app drawer in every configuration.
7. The first tap on the search bar opens the keyboard.

### Device information

- Found by reading the sources on `16-dev` (`def0a3121b`), not by testing a range of devices. None of these look device-specific; 6 is the only one with a configuration dependency (it needs a phone-sized portrait window with blur off).

### App version

16.0 Development 5.1 (`16-dev`, `def0a3121b`)

### Additional context

A branch fixing all seven is ready; PR follows and links back here.

TakuyaNarita:
### Description

Makes the icon pack and app drawer settings reach the surfaces they claim to. Six commits: three restore settings that are read by code but can no longer be set (or applied only in one of the cases they promise), and three add the settings those fixes exposed as missing — an icon pack that can be limited to the home screen, drawer search bar styling, and per-surface suggestion switches.

Fixes #7254

### Reasoning

The icon pack is applied while the icon bitmap is built, so the cache holds one bitmap per component that both the home screen and the app drawer read. Keeping the plain icons in a second cache (`DrawerIconCache`) is what lets the two surfaces differ, the same way the themed icon scope already does. It is only built once something asks for it, so nothing changes for anyone who leaves "Apply icon pack to" on its default of both surfaces.

That split is also what surfaced the drag-out bug: an app dragged out of the drawer carried the icon the drawer drew. The three places that turn a drawer item into a workspace item now fill the icon first. The cache may only be read from its own thread, so the lookup is handed to that thread and waited on with a cap — a model load can hold that thread for seconds, and a late icon is better than a frozen launcher.

The icon style page kept its settings inside the tab whose pack they belong to, so half of them were invisible depending on which tab was open. The tabs now hold only the pack grids and every setting sits below them, grouped by what it affects; the page scrolls in portrait since it no longer has to fit on one screen.

### Commits

- `50d56e78f` **fix:** Restore icon pack tint switch and themed icon pack list — both lost in ba62bdf380. `tintIconPackBackgrounds` had no switch, and `IconPackGrid` stopped honouring `isThemedIconPack`, so the themed tab listed regular packs.
- `30a5ecf79` **feat:** Allow limiting the icon pack to the home screen — adds `DrawerIconCache` and the "Apply icon pack to" setting. Drawer icons load in bulk from `LoaderTask` rather than `AllAppsList.add()`, so both paths go through the drawer cache; predictions pick their cache from the container they belong to and copy the `AppInfo` before filling its icon.
- `ae6c46920` **fix:** Make icon and drawer settings findable — the drawer page links to icon style, the icon style page shows every setting on every tab, a Lawnicons link appears when no supported pack is installed, the dock's icon count says it is shared with the home screen grid, and the drawer's suggested apps switch is back on the predictions page.
- `1d415cc51` **fix:** Give the home screen its icon when an app leaves the drawer — plus the update pass and freshness key the drawer cache was missing.
- `cc0903c63` **feat:** Style the drawer search bar, and open the keyboard on the first tap — corner radius, background opacity and stroke become settings with a preview, defaults unchanged. The keyboard retry watches window insets rather than what the input method returns, which is whether the request was accepted, not whether the keyboard came up.
- `adaaf748b` **feat:** Suggestions per surface, an empty dock, and a drawer that dims — separate switches for the drawer row and the dock slots under the main predictions switch; the dock accepts zero rows; drawer background opacity applies in every case, not just the sheet-without-blur one.

### Testing

**Icon pack tint and themed list**
1. Settings → Home screen → Icon style → Icon pack tab → "Tint with accent color" is present and takes effect.
2. Themed icons tab → only packs answering `app.lawnchair.icons.THEMED_ICON` are listed. With none installed, a link to Lawnicons appears.

**Icon pack scope**
3. Icon style → "Apply icon pack to" → **Home screen and app drawer** (default): both surfaces use the pack, as before.
4. Switch to **Home screen**: the drawer falls back to plain icons, the home screen keeps the pack. Switch back and both return.
5. With either setting, drag an app out of the drawer onto the home screen — the dropped icon is the home screen's icon immediately, no restart needed.
6. Change an app's icon (or install/uninstall one) and confirm the drawer picks it up without a launcher restart.

**Settings visibility**
7. Icon style → switch between tabs → every setting below the grid stays visible. In portrait the page scrolls.
8. App drawer page → the link to icon style opens it.

**Search bar**
9. Open the drawer, tap the search bar once → the keyboard comes up on that first tap.
10. Settings → App drawer → search bar corner radius / background opacity / stroke → the preview and the real bar follow them; defaults look as before.

**Suggestions and dock**
11. Settings → predictions → turn the drawer row and the dock slots off independently; the main switch still turns both off.
12. Dock → set rows to 0 → the icons go and the search bar stays, with no leftover gap.
13. App drawer → background opacity on a phone in portrait (blur off) → the drawer dims as the slider moves.

### Type of change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01T12aQdgNXwaNV4ejKxvcto

## Summary by CodeRabbit

* **New Features**
  * Choose whether icon packs apply to the home screen, app drawer, or both.
  * Customize the app-drawer search bar’s corner radius, opacity, border, and color.
  * Control suggested apps in the app drawer and dock independently.
  * Configure the dock with zero icon rows while keeping the search bar visible.
  * Added easier access to icon style settings and Lawnicons.
* **Bug Fixes**
  * Improved icon consistency when moving apps between the drawer and home screen.
  * App-drawer background opacity now applies consistently across display modes.
  * Improved reliability when showing the keyboard in app-drawer search.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -140,7 +140,7 @@ final def isReleaseBuild = ciBuild
 final def devReleaseName = ciBuild ? "Dev.(#${ciRunNumber})" : "Dev.(${buildCommit})"
 
 final def version = "16"
-final def releaseName = "Development 5.1"
+final def releaseName = "Development 5.2"
 final def versionDisplayName = "${version}.${isReleaseBuild ? releaseName : devReleaseName}"
 final def majorVersion = versionDisplayName.split("\\.")[0]
 
@@ -151,10 +151,10 @@ android {
     namespace "com.android.launcher3"
     defaultConfig {
         /*
-         * Lawnchair Launcher 16.0 Development 5.1
+         * Lawnchair Launcher 16.0 Development 5.2
          * see CONTRIBUTING.md#versioning-scheme
          */
-        versionCode 16_00_01_05_01
+        versionCode 16_00_01_05_02
         versionName "${versionDisplayName}"
         buildConfigField "String", "VERSION_DISPLAY_NAME", "\"${versionDisplayName}\""
         buildConfigField "String", "MAJOR_VERSION", "\"${majorVersion}\""
--- a/build.gradle
+++ b/build.gradle
@@ -140,7 +140,7 @@ final def isReleaseBuild = ciBuild
 final def devReleaseName = ciBuild ? "Dev.(#${ciRunNumber})" : "Dev.(${buildCommit})"
 
 final def version = "16"
-final def releaseName = "Development 5.2"
+final def releaseName = "Development 5.9"
 final def versionDisplayName = "${version}.${isReleaseBuild ? releaseName : devReleaseName}"
 final def majorVersion = versionDisplayName.split("\\.")[0]
 
@@ -151,10 +151,10 @@ android {
     namespace "com.android.launcher3"
     defaultConfig {
         /*
-         * Lawnchair Launcher 16.0 Development 5.2
+         * Lawnchair Launcher 16.0 Development 5.9
          * see CONTRIBUTING.md#versioning-scheme
          */
-        versionCode 16_00_01_05_02
+        versionCode 16_00_01_05_09
         versionName "${versionDisplayName}"
         buildConfigField "String", "VERSION_DISPLAY_NAME", "\"${versionDisplayName}\""
         buildConfigField "String", "MAJOR_VERSION", "\"${majorVersion}\""
--- a/lawnchair/res/values/config.xml
+++ b/lawnchair/res/values/config.xml
@@ -140,6 +140,7 @@
     <bool name="config_default_show_hidden_apps_in_search">false</bool>
     <bool name="config_default_enable_smart_hide">false</bool>
     <bool name="config_default_show_suggested_apps_at_drawer_top">true</bool>
+    <bool name="config_default_show_suggested_apps_in_dock">true</bool>
     <bool name="config_default_enable_font_selection">true</bool>
     <bool name="config_default_enable_smartspace_calendar_selection">true</bool>
     <bool name="config_default_dts2">true</bool>
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -962,6 +962,7 @@
     <string name="fuzzy_search_desc">Approximate matching for app searches</string>
 
     <string name="show_suggested_apps_at_drawer_top">Show suggested apps at the top of the drawer</string>
+    <string name="show_suggested_apps_in_dock">Show suggested apps in the dock</string>
 
     <string name="perform_wide_search_title">Device search</string>
     <string name="perform_wide_search_description">Search your phone contacts, files, and settings</string>
--- a/lawnchair/res/values/strings.xml
+++ b/lawnchair/res/values/strings.xml
@@ -392,6 +392,10 @@
     <string name="force_monochrome_label">Force monochrome</string>
     <string name="force_monochrome_description">Recolor unsupported apps with the current accent color.</string>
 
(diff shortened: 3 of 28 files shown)


TakuyaNarita:
Reopened as #7259 from my personal account (same commits, same branch). Closing this one to avoid a duplicate.

